PR #1, opened 2021-01-20
New generate reset key subcommand and minor fixes

nick-child-ibm:
This PR is to update minor bugs and minor enhancements. These include:
  -Fixing a typo of the default path to secure variables 
  -Edited compiler optimization from -Os to -O2 because -Os was causing issues on ppc (opened a gcc bug report [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98692))
  -Allow for custom man page destination during install with `make install MANDIR=<path>`

I also created an easy way to generate key reset files with `secvarctl generate reset -n <varName> -k <signerKey> -c <signerCrt>`
This update has been reflected in documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,6 @@ else
 	CRYPTO_OBJ = $(patsubst %,$(SKIBOOTOBJDIR)/%, $(_CRYPTO_OBJ))
 	OBJ += $(CRYPTO_OBJ)
 endif
-# OBJDIR=obj
-# OBJ=$(patsubst %, $(OBJDIR)/%,$(_OBJ))
 
 
 secvarctl: $(OBJ) 
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ OBJ +=$(SKIBOOT_OBJ) $(EXTRAMBEDTLS)
 
 OBJCOV = $(patsubst %.o, %.cov.o,$(OBJ))
 
-
+MANDIR=usr/local/share/
 #use STATIC=1 for static build
 STATIC = 0
 ifeq ($(STATIC),1)
@@ -72,6 +72,6 @@ secvarctl-cov: $(OBJCOV)
 install: secvarctl
 	mkdir -p $(DESTDIR)/usr/bin
 	install -m 0755 secvarctl $(DESTDIR)/usr/bin/secvarctl
-	mkdir -p $(DESTDIR)/usr/local/share/man
-	mkdir -p $(DESTDIR)/usr/local/share/man/man1
-	install -m 0755 secvarctl.1 $(DESTDIR)/usr/local/share/man/man1
+	mkdir -p $(DESTDIR)/$(MANDIR)/man
+	mkdir -p $(DESTDIR)/$(MANDIR)/man/man1
+	install -m 0755 secvarctl.1 $(DESTDIR)/$(MANDIR)/man/man1
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 #_*_MakeFile_*_
 CC = gcc 
-_CFLAGS = -Os -s -g -std=gnu99
+_CFLAGS = -O2 -s -g -std=gnu99
 LFLAGS = -lmbedtls -lmbedx509 -lmbedcrypto -Wall -Werror 
 
 _DEPEN = secvarctl.h prlog.h err.h generic.h 
--- a/README.md
+++ b/README.md
@@ -40,10 +40,12 @@ For any questions regarding secvarctl, feel free to reach out: [Nick Child](nick
      - From a hash: `$secvarctl generate h:e -h <hashAlgUsed> -i <inputHash> -o <out.esl>`  
      - From a generic file (hash done internally) : `$secvarctl generate f:e -h <hashAlgToUse> -i <inputFile> -o <out.esl>`   
    + Signed Auth File (EXPERIMENTAL):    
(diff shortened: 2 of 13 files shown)


secvarctl.c
```diff
@@ -123,7 +123,7 @@ static struct backend *getBackend()
 		prlog(PR_WARNING, "WARNING!! Could not extract data from %s , assuming platform does not support secure variables\n", secVarFormatLocation);
 		goto out;
 	}
-	//loop through all known backends
+	// loop through all known backends
```
richteer, line 126: This commit does not need to be in the history. If you want to keep this comment change, change the description instead to reflect what it is doing.

README.md
```diff
@@ -191,6 +196,8 @@ For any questions regarding secvarctl, feel free to reach out: [Nick Child](nick
 		A PKCS7 and Auth file can be signed with several signers by adding more ' -k <privKey> -c <cert>' pairs. 
 		Additionaly, when generating an Auth file the secure variable name must be given as -n <keyName> because it is included in the  message digest. 
 		When using the input type '[f]ile' it will be assumed to be a text file and if output file is '[e]sl', '[p]kcs7' or '[a]uth' it will be hashed according to <hashAlg> (default SHA256). 
+		To create a variable reset file (one that will remove the current contents of a variable), replace '<inputFormat>:<outputFormat>' with 'reset' and
+		supply a variable name, public and private signer files and an output file with '-n <keyName> -k <privKey> -c <crtFile> -o <outFile>'
 		GENERATION OF PKCS7 AND AUTH FILES ARE IN EXPERIMENTAL DEVELEPOMENT PHASE. THEY HAVE NOT BEEN THOROUGHLY TESTED YET.
```
richteer, line 203: Commit message should include a little more about what is being updated about this file. It seems like it is adding documentation for the `reset` command? Not sure what changed about the blob at the top, probably whitespace?

Makefile
```diff
@@ -46,8 +46,6 @@ else
 	CRYPTO_OBJ = $(patsubst %,$(SKIBOOTOBJDIR)/%, $(_CRYPTO_OBJ))
 	OBJ += $(CRYPTO_OBJ)
 endif
-# OBJDIR=obj
-# OBJ=$(patsubst %, $(OBJDIR)/%,$(_OBJ))
```
richteer, line 49: Again, please don't include Jenkins-related commits in the history

backends/edk2-compat/include/edk2-svc.h
```diff
@@ -13,7 +13,7 @@
 #define CERT_BUFFER_SIZE        2048
 
 #ifndef SECVARPATH
-#define SECVARPATH "/sys/firmware/secvar/vars"
+#define SECVARPATH "/sys/firmware/secvar/vars/"
 #endif
```
richteer, line 18: This short commit message is fine, since it's literally a one character change.

Makefile
```diff
@@ -1,6 +1,6 @@
 #_*_MakeFile_*_
 CC = gcc 
-_CFLAGS = -Os -s -g -std=gnu99
+_CFLAGS = -O2 -s -g -std=gnu99
```
richteer, line 3: This should be its own commit, with its own description explaining why this change was made. If this needed to be bisected by someone else, knowing why this change was made may make a big difference, which is why it shouldn't be lumped in with another change.

secvarctl.c
erichte-ibm, line 126: Don't include Jenkins-related commits in the history here. If you want to keep this change, have the commit message reflect what is actually changing in the commit.

README.md
erichte-ibm, line 203: General commit message comment: Specify what updated about the readme. Looks like the reset command was added? Something else changed about the whitespace i guess in that other list?

Makefile
```diff
@@ -46,8 +46,6 @@ else
 	CRYPTO_OBJ = $(patsubst %,$(SKIBOOTOBJDIR)/%, $(_CRYPTO_OBJ))
 	OBJ += $(CRYPTO_OBJ)
 endif
-# OBJDIR=obj
-# OBJ=$(patsubst %, $(OBJDIR)/%,$(_OBJ))
```
erichte-ibm, line 50: Again, please try not to include Jenkins commits in the final commit history.

backends/edk2-compat/include/edk2-svc.h
```diff
@@ -13,7 +13,7 @@
 #define CERT_BUFFER_SIZE        2048
 
 #ifndef SECVARPATH
-#define SECVARPATH "/sys/firmware/secvar/vars"
+#define SECVARPATH "/sys/firmware/secvar/vars/"
```
erichte-ibm, line 16: Note: this short commit message is fine, since it's literally a 1 character typo fix that isn't largely consequential.

Makefile
erichte-ibm, line 3: This one-character change however, is very consequential. This should be its own commit, with a longer detailed message explaining why this change is mad. Consider someone else trying to debug an issue, and running bisect -- this may be important information to someone in the future using an old GCC wondering why they are seeing the issue again, etc.

erichte-ibm (changes requested):
Aside from the Jenkins commits, the rest largely looks fine.

The major things I would like changed before merge:
 - Remove the Jenkins commits (or merge them and call it a "formatting fix" instead)
 - Split out the optimization change into its own commit

In the future, I'd prefer some of the larger commits separated out -- try not to merge fixes with features. Fixes should be their own commits, even if they are small, as they provide documentation in the git history on when a bug was addressed.

PR #2, opened 2021-01-26
Add functionality for external signatures

nick-child-ibm:
This PR includes the implementation, documentation, and the testing for secvarctl ability to accept externally signed data. Rather than only supporting signatures computed internally. This will let secvarctl produce the presigned digest, allowing the user to use an external signing framework for the actual signature generation. From there, secvarctl can use this raw signed data when generating an auth or PKCS7 file. The process will look like this:
   $ secvarctl generate c:x  -n \<varName> -t \<timestamp> -i \<newCrt> -o \<digest.bin>
   $<user generates the signature using another tool, resulting in file \<signature.bin>
   $secvarctl generate c:a -n \<varName> -t \<timestamp> -i \<newCrt> -o \<update.auth> -s \<signature.bin> -c \<signingCrt.pem>
NOTE: The timestamp and varName must be the exact same for both secvarctl commands

### DIFF
--- a/extraMbedtls/generate-pkcs7.c
+++ b/extraMbedtls/generate-pkcs7.c
@@ -49,13 +49,14 @@ extern int verbose;
 typedef struct PKCS7Info {
 	unsigned char **crts; // signing crt DER
 	size_t *crtSizes;
-	unsigned char **keys; // signing key DER
+	unsigned char **keys; // signing key DER or signatures (depends on alreadySignedFlag)
 	size_t *keySizes;
 	int keyPairs;
 	const unsigned char *newData; 
 	int newDataSize;
 	mbedtls_md_type_t hashFunct;
-	const char * hashFunctOID;  
+	const char * hashFunctOID; 
+	int alreadySignedFlag; //if this is 1 then then PKCS7Info.keys contains signatures, if 0 then contains siging key in DER format 
 
 } PKCS7Info;
 #endif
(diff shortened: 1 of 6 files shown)


extraMbedtls/generate-pkcs7.c
```diff
@@ -816,7 +817,7 @@ int to_pkcs7_already_signed_data(unsigned char **pkcs7, size_t *pkcs7Size, const
 		if (sigs[i]) free(sigs[i]);
 	}
 	if (sigs) free (sigs);
-
+	if (sig_sizes) free(sig_sizes);
```
erichte-ibm, line 820: Are these fixes to a bug/leak introduced in the first commit of this PR? If so you can just squash this commit into that one, no need to have a separate commit for it.

If this is fixing an unrelated bug, ignore this.

extraMbedtls/generate-pkcs7.c
```diff
@@ -49,13 +49,14 @@ extern int verbose;
 typedef struct PKCS7Info {
 	unsigned char **crts; // signing crt DER
 	size_t *crtSizes;
-	unsigned char **keys; // signing key DER
+	unsigned char **keys; // signing key DER or signatures (depends on alreadySignedFlag)
```
erichte-ibm, line 52: So, I'm not fully up on the pkcs7 implementation for mbedtls, but make sure if you are adding features to the pkcs7 generation code for new features in secvarctl, that those changes make sense for mbedtls as well. 

It may be suggested that rolling pre-signed data into a pkcs7 blob should be handled in glue code outside of the library (i.e. in secvarctl) instead of in the library itself. Not suggest you change this now, but more of a warning. Maybe look at other implementations of similar algos/specs/etc and see if there's precedence for this?

erichte-ibm:
Looks good to me, I'm marking as approved, but will wait for an answer on the squash commit question before merging.

erichte-ibm (approved):

PR #3, opened 2021-02-18
CMakeLists.txt: Add Cmake build options

nick-child-ibm:
This commit adds CMakeLists.txt. The user will now have the option to build with Cmake or Make. The Cmake build will have the same options as Make. Static, Debug, Coverage, Custom Mbedtls and  non crypto builds can all be generated with STATIC=On , CMAKE_BUILD_TYPE=Debug, CMAKE_BUILD_TYPE=Coverage, CUSTOM_MBEDTLS=<path>, and NO_CRYPTO=On respectively. This commit also removes the -g option from the Makefile's CFLAGS because it had no impact with the following -s and -O2 flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,94 @@
+cmake_minimum_required( VERSION 3.12 )
+project( secvarctl )
+
+set(CMAKE_C_COMPILER gcc)
+#sources/dependencies for secvarctl
+set( DEPEN secvarctl.h prlog.h err.h generic.h )
+set( DEPDIR include/ )
+list( TRANSFORM DEPEN PREPEND ${DEPDIR} )
+set( SRC secvarctl.c generic.c )
+
+#sources/dependencies for edk2 backend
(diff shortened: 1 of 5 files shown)


sharkcz:
I think you shouldn't be using "-s" by default as it will affect the ability to produce debugging information. IMO it's better to let the stripping be a separate (optional) operation and let  `-g -O2` be a good default.

nick-child-ibm:
> I think you shouldn't be using "-s" by default as it will affect the ability to produce debugging information. IMO it's better to let the stripping be a separate (optional) operation and let `-g -O2` be a good default.

Oh hey Dan! Thanks for looking over my PR! Funny you say that because while I was waiting for this to be reviewed, I tried to generate an RPM  with cmake and the last issue I had to fix was adding the debug info. I only wish I saw this sooner. I think you are right about the `-s` flag, I am going to add an option for this, something like `STRIP=On` . Thanks again!

PR #5, opened 2021-02-24
Add Apache license to all files

nick-child-ibm:
This commit prepends a SPDX license tag to all source files. This license is Apache 2.0. The full license can be found in the LICENSE file. All test data also needed to be regenerated since the python scripts had been edited.


erichte-ibm:
Few things to fix before this can be merged:
1. Keep any files that already had license statements as they were (e.g. any files pulled from skiboot. If Modifications were made, put a comment at the top noting what had to be changed. Work item for the future: not require any modifications at all)
2. Revert the test data as it was
3. Remove any makefile dependency rule for generating the test data -- we should not be regenerating the static test data any time a script is updated.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -6,10 +6,10 @@
 #include <ctype.h> // for isspace
 #include <mbedtls/md.h>     /* generic interface */
 #include <mbedtls/platform.h> /*mbedtls functions*/
-#include "../../extraMbedtls/include/pkcs7.h" // for PKCS7 OID
-#include "include/endian.h"
+#include "../../external/extraMbedtls/include/pkcs7.h" // for PKCS7 OID
+#include "../../external/skiboot/include/endian.h"
```
erichte-ibm, line 12: Maybe add `-I external/` to the compiler flags, so all of these can be changed to `#include <skiboot/include/endian.h>`?

Alternatively, `-I .` (if not already set) would work so that all headers are preceded with `external/`, to make it clear its coming from somewhere else.

Either way, these long relative paths can get a little noisy

nick-child-ibm: I kind of like them as is. I think for clarity it is nice to see what headers are part of the package and which ones are expected to be on the system already. Maybe, I am just being lazy but let me know if this needs to be changed

erichte-ibm: Leave it for now, but this should probably get cleaned up. I think it might help keep the external code completely unchanged, since we wouldn't have to modify header paths.

PR #6, opened 2021-03-01
Enable -Wall and -Werror

daxtens:
They were being added in the link flags rather than the compile
flags, so they were not having any effect at all.

I've tried to propagate the `unsigned char` idiom from mbedtls outwards, but not into the skiboot parts just yet.

Signed-off-by: Daniel Axtens <dja@axtens.net>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,8 +93,7 @@ set( CMAKE_C_FLAGS_DEBUG       "-O0 -g3" )
 set( CMAKE_C_FLAGS_COVERAGE    "-O0 -g3 -fprofile-arcs -ftest-coverage" )
 
 
-#add linker flags ( -lmbedtls -lmbedcrypto -lmbedx509 not needed since linked above )
-target_link_options( secvarctl PUBLIC -Wall -Werror )
+target_compile_options( secvarctl PRIVATE -Wall -Werror )
 
 
 
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@
 # Copyright 2021 IBM Corp.
 #_*_MakeFile_*_
 CC = gcc 
-_CFLAGS = -s -O2 -std=gnu99
-LFLAGS = -lmbedtls -lmbedx509 -lmbedcrypto -Wall -Werror 
+_CFLAGS = -s -O2 -std=gnu99 -Wall -Werror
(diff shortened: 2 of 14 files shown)


sharkcz:
It's not recommended to use `-Werror` for distribution builds. It easily breaks automated rebuilds when a new toolchain version introduces new warnings.

daxtens:
Maybe we could put it in all the debug builds for CMake then, but not sure
what the best way to gate that in make would be - maybe another environment
variable?

Nick, what do you think?

>

nick-child-ibm:
First, I want to thank you for the PR, this is a pretty big one and I thank you for noticing this as well as implementing all the fixes for it. 

Since distro builds use the "Release" build type with Cmake, I think we could get away with adding the flags to the "Debug" and "Coverage" build type only. As for Make, I think we should add the flags only if the coverage build type is requested. If the user wants to build with -werror and -wall for the standard build, they can set the CFLAGS variable manually (it was intentionally left blank for these circumstances). I think this works best because it keeps the default builds for both cmake and make as similar as possible. Let me know if this works for everyone and I can commit the changes.

erichte-ibm:
> Since distro builds use the "Release" build type with Cmake, I think we could get away with adding the flags to the "Debug" and "Coverage" build type only.

Sounds good to me, but...

> As for Make, I think we should add the flags only if the coverage build type is requested. If the user wants to build with -werror and -wall for the standard build, they can set the CFLAGS variable manually (it was intentionally left blank for these circumstances).

Users aren't likely to be building this from source, I'd rather the default build from a fresh clone have `-Wall` and `-Werror` enabled. CI should probably also expect those to be enabled too.

I'm thinking:
 - set `-Wall`, `-Werror` on `Debug` at minimum
 - set `Debug` to be the default build
 - have the specfile build on `Release` (I think it just builds the default right now, which it should probably always specify `Release`)

nick-child-ibm:
>     * have the specfile build on `Release` (I think it just builds the default right now, which it should probably always specify `Release`)
@sharkcz  Is it okay to specify the build type in the spec file? Then we can set default Cmake Build type to "Debug". Add `-werror -wall` to CMake "Debug" and "Coverage" build types and the for all builds with Make. For distro builds we will build with Cmake and specify we want build type "Release".

PR #7, opened 2021-03-11
CMakeLists.txt: change default build type to Debug

nick-child-ibm:
This commit will change the default build type to debug. Both debug and coverage have been given additional flags: -Wall and -Werror. The reasoning is that most of the time that the project is built from source, it will be to incorporate/test changes or enhancements. This will encourage all warnings to be fixed before commiting a change. If building for general use, then the Release build type can be used (these flags have not been added here for this reason)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,18 +82,16 @@ endif (  )
 target_link_libraries( secvarctl ${MBEDTLS} ${MBEDX509} ${MBEDCRYPTO} ${PTHREAD} ) 
 
 #set default build type to release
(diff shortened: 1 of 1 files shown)


daxtens:
:ship: Ship It :ship:

I'm not sure if you have a specfile for building on RPM distros - I don't see it in the repo, I don't know if there's magic for Cmake projects, but as long as Release gets set for your production builds, :ship: .

sharkcz:
I think this is OK for distros, the `%cmake` macro expects a Release build to be the default, but this can be easily overriden.

PR #8, opened 2021-03-11
Use argp for argument parsing

nick-child-ibm:
This PR is to replace the current argument parsing scheme with one that incorporates argp. Argp is a package from the GNU C standard library, it greatly reduces the need for argument parsing functions and autogenerates help and usage messages. This PR contains commits that add the argp implementation to each of the subcommands {'read', 'write', 'verify', 'validate', 'generate' }. This PR does not alter or remove any of the current command line arguments. However, it does add long flags for almost all of the options (ex: `-v` can also be specified with `--verbose`). It is very likely that another commit may be added in the future to trim some of the added source code. For example, every subcommand individually creates the 'program name' argument by concatenating the first and second command line element. This method was just favorable when I was editing and testing every command one by one. Now that they all work and use the same logic, I will likely move this operation to the `main` function to avoid duplicate code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,15 @@ For any questions regarding secvarctl, feel free to reach out: [Nick Child](nick
 		-v , verbose, gives process info
 		-n <varName> , name of secure boot variable, used when generating an auth file, PKCS7, or when the input file contains hashed data rather than x509 (use '-n dbx'), current <varName> are: {'PK','KEK','db','dbx'}
 		-f force generation, skips validation of input file, assumes format to be correct
-		-t <time> , where time is of the format 'y-m-d h:m:s'. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used
+		-t <time> , where <time> is of the format described below. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used, all times are in UTC
+                    format of <time> = 'YYYY-MM-DDThh:mm:ss' where:
+                        - 'YYYY' four-digit year
+                        - 'MM' two-digit month (01=January, etc.)
+                        - 'DD' two-digit day of month (01 through 31)
+                        - 'T' appears literally
+                        - 'hh' two digits of hour (00 through 23) (am/pm NOT allowed)
+                        - 'mm' two digits of minute (00 through 59)
+                        - 'ss' two digits of second (00 through 59)
 		-h <hashAlg> hash function, used when output or input format is [h]ash, current <hashAlg> are : {'SHA256', 'SHA224', 'SHA1', 'SHA384', 'SHA512'}
 		-k <privKey> , private key, used when generating [p]kcs7 or [a]uth file
 		-c <certFile> , x509 certificate (PEM), used when generating [p]kcs7 or [a]uth file
--- a/backends/edk2-compat/edk2-svc-generate.c
+++ b/backends/edk2-compat/edk2-svc-generate.c
@@ -236,12 +236,20 @@ static int parse_opt(int key, char *arg, struct argp_state *state)
             }
             args->pkcs7_gen_meth = W_PRIVATE_KEYS;
 			args->signKeyCount++;
-			args->signKeys = realloc(args->signKeys, args->signKeyCount * sizeof(char*));
+			rc = reallocArray((void **)&args->signKeys, args->signKeyCount, sizeof(*args->signKeys));
+			if (rc) {
+            	prlog(PR_ERR, "Failed to realloc private key (-k <>) array\n");
+            	break;
+            }
 			args->signKeys[args->signKeyCount - 1] = arg;
 			break;
 		case 'c':
 			args->signCertCount++;
-			args->signCerts = realloc(args->signCerts, args->signCertCount * sizeof(char*));
+			rc = reallocArray((void **)&args->signCerts, args->signCertCount, sizeof(*args->signCerts));
+			if (rc) {
+            	prlog(PR_ERR, "Failed to realloc certificate (-c <>) array\n");
+            	break;
+            }
 			args->signCerts[args->signCertCount - 1] = arg;
 			break;
 		case 'f':
@@ -259,7 +267,11 @@ static int parse_opt(int key, char *arg, struct argp_state *state)
             }
             args->pkcs7_gen_meth = W_EXTERNAL_GEN_SIG;
             args->signKeyCount++;
-            args->signKeys = realloc(args->signKeys, args->signKeyCount * sizeof(char*));
+            rc = reallocArray((void **)&args->signKeys, args->signKeyCount, sizeof(*args->signKeys));
+            if (rc) {
+            	prlog(PR_ERR, "Failed to realloc signature (-s <>) array\n");
+            	break;
+            }
             args->signKeys[args->signKeyCount - 1] = arg;
 			break;
 		case 'i':
--- a/backends/edk2-compat/edk2-svc-generate.c
+++ b/backends/edk2-compat/edk2-svc-generate.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <time.h> // for timestamp
 #include <ctype.h> // for isspace
+#include <argp.h>
 #include <mbedtls/md.h>     /* generic interface */
 #include <mbedtls/platform.h> /*mbedtls functions*/
 #include "../../external/extraMbedtls/include/pkcs7.h" // for PKCS7 OID
@@ -14,11 +15,8 @@
 #include "../../external/skiboot/include/edk2-compat-process.h" // work on factoring this out
 
 
-
-
-
 struct Arguments {
-    //the alreadySignedFlag is to determine if signKeys stores a private key file(0) or signed data (1)
+    // the alreadySignedFlag is to determine if signKeys stores a private key file(0) or signed data (1)
 	int helpFlag, inpValid, signKeyCount, signCertCount, alreadySignedFlag;
 	const char *inFile, *outFile, 
 	**signCerts, **signKeys,
@@ -26,8 +24,8 @@ struct Arguments {
 	char **currentVars;
 	struct efi_time *time;
 }; 
-static int parseArgs(int argc, char *argv[], struct Arguments *args);
 
+static int parse_opt(int key, char *arg, struct argp_state *state);
 static int generateHash(const unsigned char* data, size_t size, struct Arguments *args, const struct hash_funct *alg, unsigned char** outHash, size_t* outHashSize);
 static int validateHashAndAlg(size_t size, const struct hash_funct *alg);
 static int toESL(const unsigned char* data, size_t size, const uuid_t guid, unsigned char** outESL, size_t* outESLSize);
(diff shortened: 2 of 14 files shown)


daxtens (approved):
LGTM. :+1:

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -232,200 +206,148 @@ int performGenerateCommand(int argc,char* argv[])
...
-				goto out;
-			}
-			else if (!strcmp(argv[i], "--help")) {
-				help();
+		}
+		else if (state->next < state->argc)
+			if (strncmp("--u", state->argv[state->next], strlen("--u")) == 0 
+				|| strncmp("--h", state->argv[state->next], strlen("--h")) == 0
+				|| strncmp("-?", state->argv[state->next], strlen("-?")) == 0)
 				args->helpFlag = 1;
```
daxtens, line 239: man, again, what a bummer.

I think we can make it a bit simpler by defining `--help` and `--usage` ourselves and have them invoke the functions argp provides for printing help and usage. (I just got this from skimming argp.h.)

There's a trick needed for having `usage` with just a long option which I got from https://stackoverflow.com/questions/47727755/gnu-argp-how-to-parse-option-with-only-long-name.

This allows us to clean things up quite a bit, and AFAICT the only downside is that `--usage` gets sorted before `--verbose` in the help messages. I think we can live with that.

See the attached diff.

[generate.diff.txt](https://github.com/open-power/secvarctl/files/6243767/generate.diff.txt)

If this approach seems good to you (and it's possible I missed some other subtlety) then it should apply to the other subcommands as well.

daxtens: (perhaps it's conventional for --help to go to stdout, you could check the argp source to be sure.)

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -139,44 +53,103 @@ int performGenerateCommand(int argc,char* argv[])
 	size_t outBuffSize, size;
 	struct hash_funct *hashFunction;
 	unsigned char *buff = NULL, *outBuff = NULL;
+	char *programName = NULL;
 	struct Arguments args = {	
 		.helpFlag = 0, .inpValid = 0, .signKeyCount = 0, .signCertCount = 0, .alreadySignedFlag = 2,
```
daxtens, line 58: I got super confused by the meaning of `alreadySignedFlag` later on in this file. A good followup patch would be to replace it with an enum of some sort.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -232,200 +206,148 @@ int performGenerateCommand(int argc,char* argv[])
...
-					args->outFile = argv[i];
-				}
+            args->alreadySignedFlag = 0;
+			args->signKeyCount++;
+			args->signKeys = realloc(args->signKeys, args->signKeyCount * sizeof(char*));
+			args->signKeys[args->signKeyCount - 1] = arg;
+			break;
+		case 'c':
+			args->signCertCount++;
+			args->signCerts = realloc(args->signCerts, args->signCertCount * sizeof(char*));
```
daxtens, line 257: Sadly, realloc can fail and return NULL. Worse, cleanup after realloc failure is _intensely_ annoying, as it returns NULL but does *not* free the memory you passed in. (See `man realloc`.)

There's also the (**entirely** theoretical IMO) issue that the multiplication here might overflow. `reallocarray` would handle that for you but that's another gnu extension.

not sure how best to get around this, I'd say probably a wrapper function but I'm not fussy.

(btw, it occured to me that maybe if you have gcc10 you can get the compiler to find stuff like this for you: try `-fanalyzer` per https://developers.redhat.com/blog/2020/03/26/static-analysis-in-gcc-10/ Currently it finds heaps of bugs, none of which I have triaged, but it looks like at least some are in your arg parser code, so maybe worth a look at some point, but I'd advise against fixing unrelated issues in this series, save it for a followup.)

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -232,200 +206,148 @@ int performGenerateCommand(int argc,char* argv[])
...
-				if (i + 1 >= argc || argv[i + 1][0] == '-') {
-					prlog(PR_ERR, "ERROR: Incorrect flag '-n', see usage...\n");
-					rc = ARG_PARSE_FAIL;
-					goto out;
-				}
-				else {
-					i++;
-					args->varName = argv[i];
-				}
+			//make sure -t flag has two strings <date> <time>
```
daxtens, line 297: I'm pretty uncomfortable with this as an CLI design decision, I think -t should take 1 string "<date> <time>". But @erichte-ibm can weigh in.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -469,8 +391,7 @@ static int getOutputData(const unsigned char *buff, size_t size, struct Argument
 			rc = generateESL(buff, size, args, hashFunction, outBuff, outBuffSize);
 			break;
 		default:
-			prlog(PR_ERR, "ERROR: Unkown output format %s , see usage below...\n", args->outForm);
-			usage();
```
daxtens, line 473: You might be able to use `argp_help` here to print the usage message? Or have the new error message suggest trying `--help` or `--usage`...

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -9,14 +9,13 @@
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
+#include <argp.h>
```
daxtens, line 12: Many of my previous comments re generate apply here; I'll recheck this file (and the other subcommands) after we conclude on them.

daxtens (commented):
This is a big step forward. I think with a little extra work we can make it even cleaner.

I think there should either be one big commit moving all of these to argp, or 1 commit per file. (I think I lean towards 1 big commit, but I don't really mind.)

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -139,44 +53,103 @@ int performGenerateCommand(int argc,char* argv[])
...
+		prlog(PR_ERR, "ERROR: failed to allocate memory\n");
+		return ALLOC_FAIL;
+	}
+	programName = strcpy(programName, argv[0]);
+	programName = strcat(programName, " ");
+	programName = strcat(programName, argv[1]);
+	argc--;
+	argv++;
+	argv[0] = programName;
+
```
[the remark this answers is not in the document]
nick-child-ibm: So you're saying have the help message for the specific secvarctl command only display the subcommand (not the `secvarctl`  part). Similar to the output of something like `openssl x509 --help` , where the usage is listed as `Usage: x509 [options]` . I am fine with doing this, as long as everyone thinks it maintains clarity. Please confirm and I can make updates.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -139,44 +53,103 @@ int performGenerateCommand(int argc,char* argv[])
...
+	struct argp_option options[] = 
+	{
+		{"var_name", 'n', "VAR_NAME", 0 , "name of a secure boot variable, used when generating an PKCS7/Auth file."
+										" Also, when an ESL or Auth file contains hashed data use '-n dbx'."
```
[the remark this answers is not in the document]
nick-child-ibm: I think I am going to add a format PR after this and the openssl PR get merged. It would add the format file and make the necessary format edits

backends/edk2-compat/edk2-svc-generate.c
nick-child-ibm: Added a new commit for these edits. I also figured out how to do the ordering of the arguments in the help message so that `usage` and `help` are together. Thanks a bunch for figuring this out. It looks a lot cleaner

backends/edk2-compat/edk2-svc-generate.c
nick-child-ibm: Good idea, I am going to through that into a new commit in this PR.

backends/edk2-compat/edk2-svc-generate.c
nick-child-ibm: Thanks for pointing this out. Interesting realloc can return NULL without freeing. In response to this, I made a new commit that uses a new function to to realloc the array. I will add to this PR (sorry!) for sake of less PR's and getting things moving.

nick-child-ibm: As for using `-fanalyzer` , I have added updating my compiler and using the flag to the secvarctl trello board.

backends/edk2-compat/edk2-svc-generate.c
nick-child-ibm: I will wait for @erichte-ibm  to chime in. I have no strong opinions on this but would prefer to get things moving so we can make RHEL 8.5 .

backends/edk2-compat/edk2-svc-generate.c
nick-child-ibm: I do not think we can use any of the argp functions without supplying it with an ` struct argp_state *state` which we no longer have in these functions. However, I have added a `use  '--help' for more info`.

backends/edk2-compat/edk2-svc-generate.c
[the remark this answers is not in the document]
daxtens: Sorry, I wasn't clear.

I think it would be fine to print only the subcommand, but I think what you're doing now is also fine. It's not worth getting hung up on; we have bigger fish to fry.

What I want to do in the future is have argument parsing go in 2 stages.

Stage 1 is parsing global options. So here we parse everything up to the subcommand. Usually that will be nothing, but it might be something like `--verbose` or `--help`. Stage 2 is parsing the subcommand arguments, that will specialise by subcommand.

Whether or not that actually makes this particular piece of code simpler in the end I'm not sure, I just wanted to flag this as something I'd want to change later.

backends/edk2-compat/edk2-svc-generate.c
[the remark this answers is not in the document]
daxtens: works for me.

backends/edk2-compat/edk2-svc-generate.c
daxtens: Cool, I'll check that out. yes, realloc has an 'interesting' calling convention. I'll ping you internally about the trello board, I'd be interested in seeing what your grand plans are :)

backends/edk2-compat/edk2-svc-generate.c
daxtens: oh right. No worries.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -16,15 +16,24 @@
 
 
 struct Arguments {
-    // the alreadySignedFlag is to determine if signKeys stores a private key file(0) or signed data (1)
-	int helpFlag, inpValid, signKeyCount, signCertCount, alreadySignedFlag;
+    // the pkcs7_gen_meth is to determine if signKeys stores a private key file(0) or signed data (1)
+	int helpFlag, inpValid, signKeyCount, signCertCount, pkcs7_gen_meth;
```
daxtens, line 20: should pkcs7_gen_meth have type `enum pkcs7_generation_method`?

daxtens (commented):
Apart from the type of `pkcs7_gen_meth`, "edk2-svc-generate: Make the two PKCS7 generation methods..." lgtm.

generic.c
```diff
@@ -164,3 +164,33 @@ int createFile(const char * file, const char * buff, size_t size)
...
 }
 
+/*
+ *returns a new pointer to a string array with one more unused index at the end 
+ *@param arr , a pointer to the string array
+ *			will be reallocated to have one more index at the end or NULL if error
+ *@param old_length , the current length of the array, the returned array will have old_length+1 memory locations
+ *@return 0 for success or ALLOC_FAIL if fail (memory will be freed in this case)
+ */
+int reallocArray(const char ***arr, int old_length)
```
daxtens, line 174: 2 comments:
 - taking old_length just so you can allocate old_length + 1 is a bit odd. Maybe just take `new_length`.
 - if you make arr a void type, you can use it for any sort of array.

generic.c
```diff
@@ -164,3 +164,33 @@ int createFile(const char * file, const char * buff, size_t size)
...
+{
+	const char **new_arr = NULL;
+	new_arr = malloc( (old_length + 1) * sizeof(char*));
```
daxtens, line 177: Using a multiplication inside a malloc is usually considered poor practice because you can end up with integer overflow bugs. See `__builtin_mul_overflow` and friends.

I'm confused that you've re-implemented realloc from the basic primitives of malloc and free, rather than wrapping realloc to handle it's quirky return values. I would have done something more like:

```
old_arr = arr;
arr = realloc(arr, new_size);
if (arr == NULL) {
  free(old_arr);
  return ALLOC_FAILED;
}
return SUCCESS;

**Review by daxtens — COMMENTED** [review]
"Add reallocArray" is a good idea, but I think the implementation is a bit baroque! 

(Also, check out `man reallocarray` which takes number of members and size of member; perhaps if you think you might realloc use non-(char **) arrays, it might be worth considering that form of parameters.)

File: backends/edk2-compat/edk2-svc-generate.c (hunk 9 above)
Comment (nick-child-ibm) [reply backends/edk2-compat/edk2-svc-generate.c:20]: Yes... Whoops

File: backends/edk2-compat/edk2-svc-generate.c (hunk 7 above)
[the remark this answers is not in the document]
Comment (erichte-ibm) [reply backends/edk2-compat/edk2-svc-generate.c:88]: Just to throw my thoughts in here, I definitely think there should be a two-stage argparse. Along with `--verbose`, `--help`, I think the backend should also be selectable prior to a subcommand, allowing for something along the lines of:
```bash
 # generates for the default/autodetect backend setting
$ secvarctl generate ...  

# generates for the edk2-compat backend explicitly
$ secvarctl --backend edk2-compat generate ...
```

Any command should probably not have to look backwards in argv. Not super familiar with argp, but can we just hardcode the program name arg if we already know what it is supposed to be? (That is, we can infer from where it is being called -- in a backend-specific command)

backends/edk2-compat/edk2-svc-write.c
```diff
@@ -31,109 +29,119 @@ static int parseArgs(int argc, char *argv[], struct Arguments *args);
...
+	programName = strcat(programName, argv[1]);
+	argc--;
+	argv++;
+	argv[0] = programName;
+
+	struct argp_option options[] = 
+	{
+		{"verbose", 'v', 0, 0, "print more verbose process information"},
+		{"force", 'f', 0, 0, "force update, skips validation of file"},
+		{"path_to_vars", 'p', "PATH" ,0, "looks for .../<var>/update file in PATH, default is " SECVARPATH},
```
erichte-ibm, line 58: Personal nit: underscores are a bit weird in argument long names, consider replacing with hyphens, or rewording to not need them? e.g. `--path-to-vars` or `--varpath`

backends/edk2-compat/edk2-svc-generate.c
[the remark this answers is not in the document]
nick-child-ibm: I agree. Perhaps add argp parsing to the secvarctl command when we are actually supporting  multiple backends? It would look a little weird now since the only presubcommand options are `--usage/help` and `-v`.

Also, yes I can hardcode it. I will do that now.

backends/edk2-compat/edk2-svc-write.c
nick-child-ibm: Good call. fixing this....

nick-child-ibm:
The above rebasing hardcodes the program name for the subcommand help/usage messages and removes any long options with an underscore (ex: `--path_to_vars` is now just `--path`)

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -370,7 +362,7 @@ void usage()
 		"\t\t\t\tdefault is " SECVARPATH "\n"
 		"VARIABLES:\n\t{'PK','KEK','db','dbx', 'TS'}\ttype one of the following to get info on that key,\n"
 		"\t\t\t\t\tNOTE does not work when -f option is present\n\n");
-}
+}*/
```
daxtens, line 365: Probably doesn't need to be commented out?

daxtens: (as in, we could just delete it)

nick-child-ibm:
I have responded to Daniel Axtens comment about removing the commented out help/usage functions from edk2-svc-read.c . Additionally, in response to internal discussion, I have implemented changes to use a better custom timestamp argument format. The new format is `-t YYYY-MM-DDThh:mm:ss` where the time is assumed to be UTC. Documentation and testing has been updated to reflect this change as well as a few changes to ensure UTC is being used when getting the current time and an updated esl_time validation.

Please take a look at line 7 of `edk2-svc-generate.c` . I was unable to use the function `strptime` without setting it. Let me know your thoughts.
Thanks

README.md
```diff
@@ -171,7 +171,15 @@ For any questions regarding secvarctl, feel free to reach out: [Nick Child](nick
 		-v , verbose, gives process info
 		-n <varName> , name of secure boot variable, used when generating an auth file, PKCS7, or when the input file contains hashed data rather than x509 (use '-n dbx'), current <varName> are: {'PK','KEK','db','dbx'}
 		-f force generation, skips validation of input file, assumes format to be correct
-		-t <time> , where time is of the format 'y-m-d h:m:s'. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used
+		-t <time> , where <time> is of the format described below. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used, all times are in UTC
+                    format of <time> = 'YYYY-MM-DDThh:mm:ss' where:
+                        - 'T' appears literally
+                        - 'YYYY' four-digit year
+                        - 'MM' two-digit month (01=January, etc.)
+                        - 'DD' two-digit day of month (01 through 31)
```
erichte-ibm, line 179: Move `T` to be after `DD`, so the bulletpoint order matches the order in the format sequence

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -293,20 +296,10 @@ static int parse_opt(int key, char *arg, struct argp_state *state)
...
-			}
-			// make sure timestamp is correct format
-			if (sscanf(arg,"%hd-%hhd-%hhd", &args->time->year, &args->time->month, &args->time->day) != 3
-				|| sscanf(state->argv[state->next], "%hhd:%hhd:%hhd", &args->time->hour, &args->time->minute, &args->time->second) != 3) {
-				prlog(PR_ERR, "ERROR: Could not parse given timestamp, make sure it is in format 'y-m-d h:m:s'\n");
-				rc = ARG_PARSE_FAIL;
-			}
-			state->next++;
+            if (parseCustomTimestamp(args->time, arg)) {
+                prlog(PR_ERR, "ERROR: Could not parse given timestamp %s, make sure it is in format <YYYY-MM-DDThh:mm:ss\n", arg);
```
erichte-ibm, line 300: Not sure what the `<` in the message is supposed to indicate.
I see the help string around a lot, might be worth factoring that out to a macro in a header somewhere so if we have to change the timestamp (which hopefully we don't...), we don't have to change this everywhere.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -327,7 +320,7 @@ static int parse_opt(int key, char *arg, struct argp_state *state)
 			else if (args->inForm == NULL || args->outForm == NULL)
 				prlog(PR_ERR, "ERROR: Incorrect '<inputFormat>:<outputFormat>', see usage...\n");
 			else if (args->time && validateTime(args->time))
-				prlog(PR_ERR, "Invalid timestamp flag -t 'y-m-d h:m:s>' , see usage...\n");
+				prlog(PR_ERR, "Invalid timestamp flag '-t YYYY-MM-DDThh:mm:ss>' , see usage...\n");
```
erichte-ibm, line 323: Same comment as above here

secvarctl.1
```diff
@@ -374,7 +374,19 @@ OPTIONAL:
...
 .PP
 .B -t 
-<time> , where time is of the format 'y-m-d h:m:s'. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used
+<time> , where <time> is of the format described below. creates a custom timestamp used when generating an auth or PKCS7 file, if not given then current time is used, all times are in UTC
+.RS 
+format of <time> = 'YYYY-MM-DDThh:mm:ss' where:
+.RS
+- 'T' appears literally
+ - 'YYYY' four-digit year
+ - 'MM' two-digit month (01=January, etc.)
```
erichte-ibm, line 383: same as the README.md

generic.c
```diff
@@ -164,3 +164,30 @@ int createFile(const char * file, const char * buff, size_t size)
 	return SUCCESS;
 }
 
+/*
+ *returns a new pointer to an array with new length
+ *@param arr , a pointer to the array, will be reallocated to have desired size or NULL if error
+ *@param new_length , the desired length of the array
```
erichte-ibm, line 170: Is this length of the array, or size of the array? This takes in `void **` and the calls to this function cast away the type. Length implies a counted number of elements in the array, which when reallocating, we would need to know the size of each element to be able to allocate an appropriate buffer size.

I would suggest making this more clear by either taking in the *size*, forcing the caller to do the math on the size, or also take in the element size as a third argument.

generic.c
```diff
@@ -164,3 +164,30 @@ int createFile(const char * file, const char * buff, size_t size)
...
+ *@return 0 for success or ALLOC_FAIL if fail (memory will be freed in this case)
+ */
+int reallocArray(void **arr, size_t new_length)
+{
+	void *old_arr;
+	size_t new_size;
+	//if realloc returns null it does not free memory so we must keep a pointer to it
+	old_arr = *arr;
+	//check if requested size is too big
+	if (__builtin_mul_overflow(new_length, sizeof(*arr), &new_size)) {
```
erichte-ibm, line 180: This will always be `void *`, so this function appears to assume the array we are reallocating has an element size of `sizeof(void *) -> 8 (probably)`

generic.c
nick-child-ibm: made this fucntion more like `calloc` with number of elements and size of each element arguments.

nick-child-ibm:
This rebase is in response to @erichte-ibm 's feedback

erichte-ibm (approved):

backends/edk2-compat/edk2-svc-generate.c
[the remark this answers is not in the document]
daxtens: I'm happy with this for now, let's revisit it later.

backends/edk2-compat/edk2-svc-generate.c
```diff
@@ -448,9 +384,9 @@ static int getOutputData(const unsigned char *buff, size_t size, struct Argument
 			rc = generateHash(buff, size, args, hashFunction, outBuff, outBuffSize);
 			break;
         case 'x':
-            //intentional flow
+            // intentional flow
```
daxtens, line 387: not a high priority, but if you use something like `/* fall through */`, gcc will recognise it and not fire a warn when compiled with `-Wimplicit-fallthrough=N`: see https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html and https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

You don't need to change this now: any comment is better than no comment. A future patch can make sure secvarctl compiles with -Wimplicit-fallthrough.

generic.c
```diff
@@ -164,3 +164,31 @@ int createFile(const char * file, const char * buff, size_t size)
...
+	void *old_arr;
+	size_t new_size;
+	//if realloc returns null it does not free memory so we must keep a pointer to it
+	old_arr = *arr;
+	//check if requested size is too big
+	if (__builtin_mul_overflow(new_length, size_each, &new_size)) {
+		prlog(PR_ERR, "ERROR: Invalid size to alloc %zd * %zd\n", new_length, size_each);
+		goto out;
+	}
+	*arr = realloc(*arr, size_each*new_length);
```
daxtens, line 185: you could just use new_size here :P but it's of no consequence!

daxtens (approved):
I haven't looked at it in particularly fine detail, but it does look like you've resolved my previous concerns and Eric is happy, so... :ship: SS Ship It :ship:

PR #9, opened 2021-03-18
Make Makeover

daxtens:
This PR simplifies and improves our make processes.

Both cmake and makefiles support automatic dependency tracking, so you don't have to manually specify every header file that a source file depends on. 

For Make, we do this with the `-MMD` argument to gcc, which creates `.d` files that list the dependencies as a Make rule. We import those files into our makefile if they exist.

CMake does this automagically, I think with it's own parser.

I also:
 * add a `DEBUG` mode for the Makefile which stops stripping symbols and adds debug code.
 * rename `LFLAGS` to the more common `LDFLAGS`
 * do misc cleanups

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,40 +4,25 @@ cmake_minimum_required( VERSION 3.12 )
 project( secvarctl C )
 
 set( CMAKE_C_COMPILER gcc )
-#sources/dependencies for secvarctl
-set( DEPEN secvarctl.h prlog.h err.h generic.h )
-set( DEPDIR include/ )
-list( TRANSFORM DEPEN PREPEND ${DEPDIR} )
+#sources for secvarctl
 set( SRC secvarctl.c generic.c )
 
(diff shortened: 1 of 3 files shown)


nick-child-ibm:
I was worried about things like the crypto library dependencies getting mixed up with this, so I rebased it onto the openssl porting branch. Everything is working nicely and no mbedtls files are included when building with openssl.
I was also worried about this effecting the rpm builds. So I ran that locally and everything still works! Nice work, this looks good to me. I am a little curious to see if @sharkcz agrees, he has much more knowledge on this than I and can hopefully confirm that this will not effect Fedora packages.

nick-child-ibm:
Also, if this ends up gettting merged after the openssl PR then I can push the changes I made locally to fix the merge conflicts.

sharkcz:
I believe it shouldn't cause any issues with the package :-)

daxtens:
I applied this directly. Thanks all.

PR #11, opened 2021-04-07
Allow secvarctl to use openssl as crypto lib

nick-child-ibm:
This PR allows for `secvarctl` to be compiled with openssl instead of mbedtls. In Make, the command is `make OPENSSL=1`, and `cmake -DOPENSSL=1` for CMake builds.  In order for `secvarctl` to properly compile with either crypto library, I had to create the `crypto_*` functions in the `crypto` directory. Using `#ifdef`'s , `secvarctl` can now make a function call to one of two implementations of the same `crypto` function (depending on which crytpo library was used during compilation). The openssl implementations have the same results as the mbedtls implementations. Therefore, from a user perspective, the difference of using mbedtls or openssl is minimal. The only functionality that is yet to be ported is the ability to generate an auth or pkcs7 using an externally generated signature. I believe this to be acceptable for now since those who will need this functionality are those who do not have access to their private keys (if using something like an HSM). 

Additionally, this PR depends on the prlog PR  [here](https://github.com/open-power/secvarctl/pull/8)  . Also, a few commits from @daxtens have been included that add a gitignore and tidy up the include paths.


CMakeLists.txt
```diff
@@ -63,6 +72,14 @@ if ( STRIP )
   string( APPEND CMAKE_C_FLAGS "-s" )
 endif(  )
 
+
+# include paths:
+#  - include/
+#  - the root directory so that we can e.g. #include "{external,backends}/..."
+include_directories(include .)
```
sharkcz, line 79: Won't this break out-of-tree builds? I mean `mkdir build; cd build; cmake ..` which is now default in Fedora. I think it should use something like `$SRCDIR` instead of the "dot" dir.

daxtens:
I don't think it will break the build, because that line comes from my
commit and I was testing with out-of-tree builds set up with that
invocation. But let me check.

>

nick-child-ibm:
These commits did not change the openssl commits. I rebased onto the argp branch again as I respond to feedback from the argp PR [here](https://github.com/open-power/secvarctl/pull/8)

nick-child-ibm:
These commits did not change the openssl commits. I rebased onto the argp branch again as I respond to feedback from the argp PR [here](https://github.com/open-power/secvarctl/pull/8)

daxtens:
I assume you're aware of this, but you haven't changed the bits in skiboot yet, so it won't fully compile without mbedtls just yet:

```
gcc   -s -O2 -std=gnu99 -I./ -Iinclude/ -Wall -Werror -DOPENSSL -c  external/skiboot//edk2-compat.c -o external/skiboot//edk2-compat.o
In file included from external/skiboot//edk2-compat.c:23:
./external/extraMbedtls/include/pkcs7.h:28:10: fatal error: mbedtls/config.h: No such file or directory
   28 | #include <mbedtls/config.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:80: external/skiboot//edk2-compat.o] Error 1
```

I'll see if I can spin you up something.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -4,241 +4,245 @@
...
 	
-// out:
-// 	dealloc_secvar(var);
+out:
+	dealloc_secvar(var);
 	
 <<<<<<< HEAD
+<<<<<<< HEAD
+=======
+>>>>>>> 83004c7... Port openssl to secvarctl read and write
```
daxtens, line 245: I think you might have had some trouble with rebasing: you're got all these conflict markers still.

nick-child-ibm: Lesson LEARNED! Always compile and test before force pushing. Yikes.... 
Thanks for pointing this out. I have never had to deal with embedded diff statements but that was an interesting experience. I am surprised I did not see any errors or warnings when cherrypicking the commits. Just when you think you finally understand git. I did get to use git bisect though for the first time to figure out where the diffs first came up, so maybe in some weird way, it was worth it.

CMakeLists.txt
nick-child-ibm: I just tried this out with cmake and building with `rpmbuild` everything seems to be working. I do think `${SRCDIR}` should be used for clarity.

daxtens:
Agree that `${SRCDIR}` would be better. Apologies, I'm not much of a Cmake
person. Feel free to change my patch, just stick your `Signed-off-by:` line
after mine.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -16,458 +16,435 @@
...
+// 		return rc;
+// 	}
+// 	free(sizePath);
+
+// 	if (size == 0) {
+// 		prlog(PR_WARNING, "Secure Variable has size of zero, (specified by size file)\n");
+// 		/*rc = INVALID_FILE;
+// 		return rc;*/
+// 	}
+
```
erichte-ibm: I'm assuming something broke when you started re-implementing the crypto, hence this giiiiant diff of half the file commented out. It's probably not worth the effort to fix now, but in the future try to avoid commits like this that comment out large swaths of code temporarily.

Instead, try to make breaking changes either incrementally, or write all the new code elsewhere and then add a commit that switches over to the new code. For example, you could have added the crypto files as a commit, but not build/use/reference them until a later commit that rewrites the commands with the new crypto functions.

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+{
+    mbedtls_x509_crt_free((mbedtls_x509_crt *)x509);
+    free(x509);
+}
+
+int crypto_convert_pem_to_der(const unsigned char *input, size_t ilen, unsigned char **output, size_t *olen)
+{
+    return convert_pem_to_der(input, ilen, output, olen);
+}
+#endif
```
erichte-ibm, line 877: Abstractions look generally sane. One thing I would suggest: consider breaking the two implementations into two different files like `crypto-openssl.c` and `crypto-mbedtls.c`, and only building the file for the crypto library in use. Long `#ifdef`s make it a bit difficult to know which "mode" you are looking at (although, I suppose you could use context clues of function names to figure it out).

crypto/crypto.h
```diff
@@ -0,0 +1,121 @@
...
+#ifdef OPENSSL
+
+#include <openssl/obj_mac.h>
+#define CRYPTO_MD_SHA1 NID_sha1
+#define CRYPTO_MD_SHA224 NID_sha224
+#define CRYPTO_MD_SHA256 NID_sha256 
+#define CRYPTO_MD_SHA384 NID_sha384
+#define CRYPTO_MD_SHA512 NID_sha512
+
+#else
```
erichte-ibm, line 13: Personal preference: maybe set a symbol for `MBEDTLS` in the Makefile/cmake, even if it is the default. I don't suspect we'll need to port to yet another crypto library, but I am a fan of avoiding vague "else"s.

crypto/crypto.h
```diff
@@ -0,0 +1,121 @@
...
+
+
+/**====================X509 Functions ====================**/
+int crypto_get_x509_der_len(void *x509);
+int crypto_get_tbs_x509_der_len(void *x509);
+int crypto_get_x509_version(void *x509);
+int crypto_get_x509_sig_len(void *x509);
+int crypto_x509_md_is_sha256(void *x509);
+int crypto_x509_oid_is_pkcs1_sha256(void *x509);
+int crypto_x509_get_pk_bit_len(void *x509);
```
erichte-ibm, line 103: Are these `void *x509` pointers usually to structs that contain their sizes, or is there a possibility these might point to some kind of arbitrary allocated memory? If mbedtls/openssl know how big their data is supposed to be, then I'm fine with this.

I think its just the `void *` that is spooking me a little bit. Could be worth defining a type for these that can be handled in the `#ifdef`s, but that might just be overkill.

nick-child-ibm: Up to you guys. I am not sure what you mean by "structs that contain their sizes", aren't all structs some fixed size? 
While I don't think that typedef-ing adds any change in the compiled code, I think it adds clarity so I will add that now. Let me know if I should do anything else with regards to this

nick-child-ibm:
The commits have been rebased onto the most recent version of the argp branch. Additionally, in response to feedback, the void pointers have been replaced with typedefs, `MBEDTLS` is now a symbol so it must be passed in by the compiler rather than being the default and the `crypto.c` file has been broken into `crypto-openssl.c` and `crypto-mbedtls.c`.

Makefile
```diff
@@ -13,19 +12,18 @@ _LDFLAGS += -s
 endif
 
 EDK2OBJDIR = backends/edk2-compat
-_EDK2_OBJ =  edk2-svc-read.o edk2-svc-write.o edk2-svc-validate.o edk2-svc-verify.o edk2-svc-generate.o
+_EDK2_OBJ =   edk2-svc-validate.o edk2-svc-read.o
+#_EDK2_OBJ =  edk2-svc-read.o edk2-svc-write.o  edk2-svc-verify.o edk2-svc-generate.o
```
daxtens: If you're not compiling these .c files into these .o files, it shouldn't be necessary to also change the .c files. Smaller diffs will make reviewing the patches much simpler.

Makefile
```diff
@@ -39,12 +37,27 @@ else
 	STATICFLAG=
 endif
 
-#use NO_CRYPTO for smaller executable but limited functionality
+#use NO_CRYPTO for smaller executable but limited
```
daxtens, line 37: ?

backends/edk2-compat/edk2-svc-read.c
daxtens: You can also use `#if 0` if you're trying to comment out huge chunks. But as Eric says, avoiding it entirely is better. Something he doesn't mention is bisect-ability - see https://wiki.samba.org/index.php/Creating_a_Samba_patch_series#Bisectability - which is something that IMO is worth preserving at least in general. In your case, I think it should be possible  for the entire program to compile with mbedtls the whole way through...

Makefile
nick-child-ibm: In response to this and your other comment, I have rebased so that these work-in-progress commits all still compile normally for mbedtls but not with openssl. When I was still porting openssl, it was either to remove/comment out all mbedtls so that I could compile and test the result. Now that I know that porting is successful, it makes sense that I changed these commits so they are backwards compatible.

Makefile
nick-child-ibm: Darnit! Thanks good catch

nick-child-ibm:
Rebased again in response to @erichte-ibm and @daxtens concerns over commenting out / not compiling large swaths of code. Previously, I had removed any code that had not yet undergone the crypto abstraction. This led to most of secvarctl being removed in some commits. The point of "bisectability" has made it clear that these commits should still compile with mbedtls. In response, those few commits where openssl porting was not finished will behave normally with building with mbedtls but will not compile at all with openssl. This is preferred because of bisectability, backwards compatability, and it results in much smaller commits.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -280,7 +279,7 @@ int getSecVar(struct secvar **var, const char* name, const char *fullPath){
 	// since we are reading from a secvar, it can be assumed it has a <var>/size file for more accurate size
 	// fullPath currently holds <path>/<var>/data we are going to take off data and add size to get the desired file
 	strcpy(sizePath, fullPath);
-	// add null terminator so strncat works
+	//add null terminator so strncat works
```
erichte-ibm: ?

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -438,28 +437,29 @@ void printESLInfo(EFI_SIGNATURE_LIST *sigList)
 	printf("\tSignature type is: %s\n", getSigType(sigList->SignatureType));
 }
 
-// prints info on x509
-int printCertInfo(mbedtls_x509_crt *x509)
+//prints info on x509
```
erichte-ibm, line 428: space after `//`

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -600,8 +600,8 @@ static int getSizeFromSizeFile(size_t *returnSize, const char* path)
 	close(fptr);
 	// turn string into base 10 int
 	*returnSize = strtol(c, NULL, 0); 
-	// strol likes to return zero if there is no conversion from string to int
-	// so we need to differentiate an error from a file that actually contains 0
+	//strol likes to return zero if there is no conversion from string to int
+	//so we need to differentiate an error from a file that actually contains 0
```
erichte-ibm: I see this seems to be a pattern. I'd recommend doing a treewide search for this and making it consistent. (`grep -re "//\w"` may help)

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -274,51 +273,55 @@ int validateAuth(const unsigned char *authBuf, size_t buflen, const char *key)
...
-	mbedtls_pkcs7_init(pkcs7);
-	rc = mbedtls_pkcs7_parse_der(cert_data, len, pkcs7);
-	if (rc != MBEDTLS_PKCS7_SIGNED_DATA) {	// if pkcs7 parsing fails, then try new signed data format 
-			prlog(PR_ERR, "ERROR: parsing pkcs7 failed mbedtls error #%04x\n", rc);
-			goto out;	
+	// pkcs7 = malloc(crypto_get_pkcs7_struct_size() );
+	// if (!pkcs7){
+	// 	prlog(PR_ERR, "ERROR: failed to allocate memory\n");
+	// 	return ALLOC_FAIL;
+	// }
```
erichte-ibm, line 285: Delete commented out large segments unless they are absolutely still needed for reference. You can always look back at the commit history, or check out an earlier commit/branch if you need to look at it again.

crypto/crypto.h
erichte-ibm: Leave it for now, we can maybe change it in the future if it's a problem.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -600,8 +589,8 @@ static int getSizeFromSizeFile(size_t *returnSize, const char* path)
 	close(fptr);
 	// turn string into base 10 int
 	*returnSize = strtol(c, NULL, 0); 
-	//strol likes to return zero if there is no conversion from string to int
-	//so we need to differentiate an error from a file that actually contains 0
+	// strol likes to return zero if there is no conversion from string to int
+	// so we need to differentiate an error from a file that actually contains 0
```
erichte-ibm, line 593: I see some of these got fixed in the future.

General comment for this file: since it looks like most of these changes were minor to `edk-svc-read.c`, can this just get merged into the previous commit?

The commit message says it ports read and write, but it appears only `-read` and `-validate`(?) have changes.

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -604,7 +604,6 @@ int parseX509(void **x509, const unsigned char *certBuf, size_t buflen)
 			prlog(PR_ERR, "ERROR: Failed to parse x509 in DER and file is not in PEM\n");
 			return CERT_FAIL;
 		}
-		printf("WAS %lu bytes now in DER is %lu bytes\n", buflen, generatedDERSize);
```
erichte-ibm: This change seems to belong in a different commit, perhaps the previous as well.

erichte-ibm (changes requested):
Looks pretty good to me! Marking as "request changes" for now, largely because it needs a rebase before it can be merged (argp merge is causing conflicts).

There are a couple of early openssl commits that can probably be squashed together since it seems they lost their original purpose (mostly the read/write commit can probably be squashed up one).

I do have some comment format complaints, but those can be addressed later as one big clean up patch -- probably going to be caught by clang-format.

backends/edk2-compat/edk2-svc-read.c
nick-child-ibm: heehee

backends/edk2-compat/edk2-svc-read.c
nick-child-ibm: Saving these for `clang-format` PR

erichte-ibm (approved):

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+}
+
+void* crypto_get_signing_cert(void *pkcs7, int cert_num)
+{
+    mbedtls_x509_crt *pkcs7_cert = NULL;
+
+    pkcs7_cert = &((struct mbedtls_pkcs7 *)pkcs7)->signed_data.certs;
+    for (int i = 0; i < cert_num && pkcs7_cert != NULL; i++)
+        pkcs7_cert = pkcs7_cert->next;
+
```
daxtens: What should happen if you ask for cert number 42 in a pkcs7 with only 3 certs? you'll get cert 3 atm, which may or may not be what you want? not sure what openssl does here, they should at least be consistent.

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+    return ((mbedtls_x509_crt *)x509)->version;
+}
+
+int crypto_x509_is_RSA(void *x509)
+{
+    int rc;
+    rc = ((mbedtls_x509_crt *)x509)->pk.pk_info->type;
+    if (rc != MBEDTLS_PK_RSA)
+        //zero is also a pk type (MBEDTLS_PK_NONE) so return generic failure if zero so it doesnt look like a success
+       return  (rc == 0 ? CERT_FAIL : rc);
```
daxtens: again, if you use a more descriptive variable name, the test will not be surprising :P

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+{
+    if (((mbedtls_x509_crt *)x509)->sig_md == MBEDTLS_MD_SHA256) 
+        return SUCCESS;
+    else
+        return CERT_FAIL;
+}
+
+int crypto_x509_oid_is_pkcs1_sha256(void *x509)
+{
+    if (strncmp((const char *)((mbedtls_x509_crt *)x509)->sig_oid.p, MBEDTLS_OID_PKCS1_SHA256, ((mbedtls_x509_crt *)x509)->sig_oid.len) )
```
daxtens: MBEDTLS_OID_CMP

daxtens: Here, if the sig_oid.len > MBEDTLS_OID_PKCS1_SHA256 len, you'll read off into random global storage space.

crypto/crypto.c
```diff
@@ -108,6 +110,113 @@ void* crypto_get_signing_cert(void *pkcs7, int cert_num)
...
+            evp_md = EVP_sha384();
+            break;
+        case CRYPTO_MD_SHA512:
+            evp_md = EVP_sha512();
+            break;
+        default:
+            prlog(PR_ERR, "ERROR: Unknown NID (%d) for MD found in PKCS7\n", md_nid);
+            rc = PKCS7_FAIL;
+            goto out;
+    }
```
daxtens: surely there must be an openssl function that can do this for you? there is for mbedtls...

crypto/crypto.c
```diff
@@ -108,6 +110,113 @@ void* crypto_get_signing_cert(void *pkcs7, int cert_num)
...
+    if (EVP_PKEY_CTX_set_signature_md(pk_ctx, EVP_sha256()) <= 0) {
+        prlog(PR_ERR, "ERROR: Failed to set signature md for pk ctx\n");
```
daxtens: should you be using evp_md here rather than hardcoding sha256?

crypto/crypto.c
daxtens: EVP_get_digestbynid is used a bit further down the file, can that be used?

external/skiboot/edk2-compat-process.c
```diff
@@ -214,23 +216,34 @@ int get_auth_descriptor2(const void *buf, const size_t buflen, void **auth_buffe
...
 	x509_buf = zalloc(CERT_BUFFER_SIZE);
-	rc = mbedtls_x509_crt_info(x509_buf, CERT_BUFFER_SIZE, "\tCRT:", &x509);
-
-	mbedtls_x509_crt_free(&x509);
+	// NICK CHILD removed direct mbedtls call, use general crypto
+	// rc = mbedtls_x509_crt_info(x509_buf, CERT_BUFFER_SIZE, "\tCRT:", &x509);
+	rc = crypto_x509_get_long_desc(x509_buf, CERT_BUFFER_SIZE, "CRT:", x509);
+	// NICK CHILD General Question: Should we be printing to debug here?
+	// it seems wierd to parse to info string then not print
+	// NICK CHILD removed direct mbedtls call, use general crypto
```
daxtens, line 244: mmm good point. not sure, will check in future when I extract this into a more generic library.

crypto/crypto.c
```diff
@@ -217,6 +218,187 @@ int crypto_pkcs7_signed_hash_verify(void *pkcs7, void *x509, unsigned char *hash
...
+    }
+    //returns 1 for success
+    rc = i2d_PKCS7_bio(out_bio, gen_pkcs7_struct);
+    if (!rc) {
+        prlog(PR_ERR, "ERROR: Failed to convert PKCS7 Struct to DER\n");
+        rc = PKCS7_FAIL;
+        goto out;
+    }
+    //get data out of BIO and into return values
+    rc = BIO_get_mem_data(out_bio, &out_bio_der);
```
daxtens: please don't use rc for return codes and sizes.

crypto/crypto.c
```diff
@@ -479,6 +722,19 @@ int crypto_pkcs7_signed_hash_verify(void *pkcs7, void *x509, unsigned char *hash
...
+    const char** crtFiles, const char** keyFiles,  int keyPairs, int hashFunct)
+{
+    return to_pkcs7_generate_signature(pkcs7, pkcs7Size, newData, newDataSize, crtFiles, keyFiles, keyPairs, hashFunct);
+}
+
+int crypto_pkcs7_generate_w_already_signed_data(unsigned char **pkcs7, size_t *pkcs7Size, const unsigned char *newData, size_t newDataSize, 
+    const char** crtFiles, const char** sigFiles,  int keyPairs, int hashFunct)
+{
+    return to_pkcs7_already_signed_data(pkcs7, pkcs7Size, newData, newDataSize, crtFiles, sigFiles, keyPairs, hashFunct);
+}
```
daxtens: is there any reason not to move the contents of those functions in to these new functions? then you can just get rid of the old ones.

daxtens (changes requested):
I have some fairly minor changes that I think should be made. Some (`MBEDTLS_OID_CMP`) are more important than others so I'm so fussed if they don't all happen at once.

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -274,51 +273,49 @@ int validateAuth(const unsigned char *authBuf, size_t buflen, const char *key)
...
-			prlog(PR_ERR, "ERROR: parsing pkcs7 failed mbedtls error #%04x\n", rc);
-			goto out;	
+	pkcs7 = crypto_pkcs7_parse_der(cert_data, len);
+	if (!pkcs7) {
+		rc = PKCS7_FAIL;
+		goto out;	
 	}
 	// make sure digest alg is sha246
-	if (memcmp(pkcs7->signed_data.digest_alg_identifiers.p, MBEDTLS_OID_DIGEST_ALG_SHA256, strlen(MBEDTLS_OID_DIGEST_ALG_SHA256) )!= 0) {
+	if (crypto_pkcs7_is_md_sha256(pkcs7) != 0) {
```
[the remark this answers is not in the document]
nick-child-ibm: Our return value for SUCCESS is 0 so I have been using that. Let me know if this is not a good idea

daxtens:
Hmm, I mean it's potentially odd because `if (SUCCESS) { printf("SUCCESS");
}` will not print anything. But it's not necessarily bad! You just have to
be really clear and consistent in your comparisons - lots of `if (rc !=
SUCCESS)` throughout so that the intent is clear.

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+
+    //if parsing made it to here then parsing was successful
+    rc = SUCCESS;
+out:
+    if (rc)
+        return NULL;
+
+    return (void *)pkcs7;
```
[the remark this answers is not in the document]
nick-child-ibm: `pkcs7` is allocated inside `d2i_PKCS7` which returns the pointer. No data is shared between `buf` and `pkcs7`, everything is copied into `pkcs7` if needed. Both need to be freed after this function.

crypto/crypto.c
```diff
@@ -0,0 +1,425 @@
...
+    PKCS7_free((PKCS7 *)pkcs7);
+}
+
+void* crypto_get_signing_cert(void *pkcs7, int cert_num)
+{
+    X509 *pkcs7_cert = NULL;
+
+    pkcs7_cert = sk_X509_value(((PKCS7 *)pkcs7)->d.sign->cert, cert_num);
+
+    return (void *)pkcs7_cert;
```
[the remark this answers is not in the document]
nick-child-ibm: Correct, the memory for the returned certificate is a field of the pkcs7 and should not be freed. When writing these functions, I worked closely with valgrind to ensure I was understanding the usage correctly.

crypto/crypto.c
nick-child-ibm: In this function, if the requested index is greater than the actual, NULL would be returned. So in your example, NULL is returned not the third certificate since the loop exits if the current cert is NULL or if we are at the desired index. Openssl also returns NULL if the desired index is out of range: `sk_TYPE_value() returns element idx in sk, where idx starts at zero. If idx is out of range then NULL is returned`

crypto/crypto.c
nick-child-ibm: The code in `to_pkcs7_already_signed_data` uses lots of static structs and functions from `generate-pkcs7.c`. Also, I plan on adding these functions to my Mbedtls PR.

crypto/crypto.c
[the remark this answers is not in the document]
nick-child-ibm: Added a NOTE in function description in `pkcs7.h` to clarify this

nick-child-ibm:
> Hmm, I mean it's potentially odd because `if (SUCCESS) { printf("SUCCESS"); }` will not print anything. But it's not necessarily bad! You just have to be really clear and consistent in your comparisons - lots of `if (rc != SUCCESS)` throughout so that the intent is clear.

I think this would be a good idea and is should be reflected across much of the `if` statements in secvarctl. We should fix all of these in a single commit one of these days. I will make a note of doing this soon.

nick-child-ibm:
The previous rebase responds to comments left by @daxtens . Thanks!

PR #13, opened 2021-05-03
Add new source code format with clang-format

nick-child-ibm:
This pull request makes edits to much of the source code to allow for a coding standard. This standard is defined in the `.clang-format` file. This file, as described by the commit message, has been borrowed from `https://github.com/torvalds/linux/blob/master/.clang-format` and edited. To reiterate the commit message, this file retains its license in the `linux/LICENSE` file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,4 +84,11 @@ install: secvarctl
 	mkdir -p $(DESTDIR)/$(MANDIR)/man1
 	install -m 0644 secvarctl.1 $(DESTDIR)/$(MANDIR)/man1
 
+#dont add all c files, extra mbedtls and skiboot should retain its own format
+CLANG_FORMAT ?= clang-format
+format:
+	cp external/linux/.clang-format .
+	$(CLANG_FORMAT) --style=file -i *.c include/*.h crypto/*.c crypto/*.h backends/*/*.c backends/*/*/*.h
+	rm .clang-format
+
 -include $(OBJ:.o=.d)
--- a/backends/edk2-compat/edk2-svc-write.c
+++ b/backends/edk2-compat/edk2-svc-write.c
@@ -5,20 +5,19 @@
 #include <stdio.h>
 #include <errno.h>
 #include <string.h>
-#include <stdlib.h>// for exit
+#include <stdlib.h> // for exit
 #include <argp.h>
 #include "secvarctl.h"
-#include "backends/edk2-compat/include/edk2-svc.h"// import last!!
+#include "backends/edk2-compat/include/edk2-svc.h" // import last!!
 
-
-static int updateSecVar(const char *var, const char *authFile, const char *path, int force);
+static int updateSecVar(const char *var, const char *authFile, const char *path,
+			int force);
 
 struct Arguments {
 	int helpFlag, inpValid;
 	const char *pathToSecVars, *varName, *inFile;
-}; 
+};
 static int parse_opt(int key, char *arg, struct argp_state *state);
-
 
 /*
  *called from main()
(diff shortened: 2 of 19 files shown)


backends/edk2-compat/edk2-svc-write.c
```diff
@@ -5,127 +5,135 @@
...
 	int rc;
-	struct Arguments args = {	
-		.helpFlag = 0, .inpValid = 0, 
-		.pathToSecVars = NULL, .inFile = NULL, .varName = NULL
-	};
+	struct Arguments args = { .helpFlag = 0,
+				  .inpValid = 0,
+				  .pathToSecVars = NULL,
+				  .inFile = NULL,
+				  .varName = NULL };
```
erichte-ibm, line 36: Hmm, not a huge fan of what the clang-format file is doing to these struct initializers, but I think I'll survive.

backends/edk2-compat/edk2-svc-write.c
```diff
@@ -5,127 +5,135 @@
...
+#include <stdlib.h> // for exit
 #include <argp.h>
 #include "secvarctl.h"
-#include "backends/edk2-compat/include/edk2-svc.h"// import last!!
+#include "backends/edk2-compat/include/edk2-svc.h" // import last!!
 
-
-static int updateSecVar(const char *var, const char *authFile, const char *path, int force);
+static int updateSecVar(const char *var, const char *authFile, const char *path,
+			int force);
```
erichte-ibm, line 14: Seeing this around a bit, I'm assuming the misalignment here is coming from github doing weird things to the format? Might need to tweak that.

Also maybe we should consider dropping, or extending the 80-character width limit, since that's a bit much

erichte-ibm (commented):
Looks fine to me. It's a clang-format overhaul, so I don't have the time or coffee to review *every* _single_ change, but nothing is offending me about the changes I see.

I do have two minor complaints with the formatter as addressed in the comments. It's not a big enough deal for me to care right now though, so we can probably leave them be unless @daxtens  has stronger opinions than mine.

backends/edk2-compat/edk2-svc-write.c
daxtens: it's not great hey. having said that, I don't care that much either; if there's a tool that'll do it automatically for me I'll tolerate quite a bit.

Makefile
```diff
@@ -84,4 +84,10 @@ install: secvarctl
 	mkdir -p $(DESTDIR)/$(MANDIR)/man1
 	install -m 0644 secvarctl.1 $(DESTDIR)/$(MANDIR)/man1
 
+#dont add all c files, extra mbedtls and skiboot should retain its own format
+format:
+	cp external/linux/.clang-format .
+	clang-format-12 --style=file -i *.c include/*.h crypto/*.c crypto/*.h backends/*/*.c backends/*/*/*.h
```
daxtens, line 90: do we need to specify clang-format-12 here? is there a reason 11 or 13 or 9 or 6 won't work?

backends/edk2-compat/edk2-svc-write.c
daxtens: I checked it out locally (`git fetch origin pull/13/head:cf`) and it looks fine to me. 3 tabs, each 8 spaces wide, is 24 characters, and `static int updateSecVar(` is also 24 characters.

Makefile
daxtens: Can I suggest this?

```
 #dont add all c files, extra mbedtls and skiboot should retain its own format
+CLANG_FORMAT ?= clang-format
 format:
        cp external/linux/.clang-format .
-       clang-format-12 --style=file -i *.c include/*.h crypto/*.c crypto/*.h backends/*/*.c backends/*/*/*.h
+       $(CLANG_FORMAT) --style=file -i *.c include/*.h crypto/*.c crypto/*.h backends/*/*.c backends/*/*/*.h
        rm .clang-format
 
 -include $(OBJ:.o=.d)
```

That allows you to override clang-format easily but doesn't require you to have clang-format-12.

daxtens: (I tested with clang-format-11, the results seem the same.)

daxtens: Once this is resolved, I am very strongly a fan of this being merged.

nick-child-ibm: Yes. I believe any clang version will work. I had to use clang-format-12 for another `.clang-format` file I was working with (it had some fields and options specific to clang-format-12 apparently). But the current `.clang-format` does not seem to have those newer fields. I just used clang-format-6 and had no issues

nick-child-ibm: I will incorporate your recommendation and push the result. Thanks!

backends/edk2-compat/edk2-svc-write.c
nick-child-ibm: Ya not seeing anything in the `.clang-format` file that we can use to our advantage on this. Only thing I see in the docs is `AfterStruct` which just puts the opening brace on a different line.

erichte-ibm: Out of curiosity, does it complain if you put a newline after the opening `{`, and before the closing `}` like so:
```
	struct Arguments args = {
				  .helpFlag = 0,
				  .inpValid = 0,
				  .pathToSecVars = NULL,
				  .inFile = NULL,
				  .varName = NULL
	};
```

I swear I saw somewhere else where it didn't complain about it, maybe it was the list of structs initialization.

PR #16, opened 2021-05-07
Move crypto functions to Skiboot directory

nick-child-ibm:
This pull request moves all of the abstracted crypto functions and files (anything in the `crypto` directory) to `external/skiboot/libstb/secvar/crypto`. The idea behind this is to have dependency between skiboot and secvarctl only flow in one direction: secvarctl depends on skiboot. Previously, there was skiboot code depending on secvarctl code and vice versa. Ideally, this will be reflected in upstream skiboot eventually. 

Also, it should be noted that these crypto files still have dependencies on secvarctl specific files (like `generic.h`, `prlog.h`, pkcs7.h`). As more skiboot is included in secvarctl, we plan on removing these skiboot-external dependencies. 

I did have a general question for reviewers:
Was it appropriate to move the files to  `external/skiboot/libstb/secvar/`? I was conflicted because there is already a directory called `crypto` in  upstream Skiboot in `libstb`. I was debating making an `abstract_lib` directory in the already existing `libstsb/crypto` directory but decided against it since these files are only being used by files in `libstb/secvar`. I am open to opinions on this though.


daxtens:
I don't mind it living in libstb/secvar at all. libstb is (at least imo) less a library and more a collection of related things. It contains mbedtls and the tss2 library, for example. All in all I'm very happy to keep this crypto abstraction layer as a secvar specific thing.

daxtens:
This looks good to me. Once you rebase it, if it passes tests I'm pretty happy for you to just merge it yourself.

daxtens:
Hmm, interesting on the clang-format. The file comes from Skiboot, which clearly has slightly different conventions. I prefer the skiboot conventions but not as much as I want the clang-format command to be the source of truth about how formatting is!

daxtens:
Thank you very much for doing this! We inch closer to being able to spin out libstb-secvar as its own library!

PR #17, opened 2021-05-11
Further syncing with skiboot

daxtens:
This series:

 - brings in compiler.h, which is pretty handy.
 - brings in edk2.h from skiboot, as it will be if my patches to the list are accepted.
 - clean up the pragma pack warnings now that edk2.h doesn't use them
 - exports some functions in skiboot and makes them accessible to secvarctl. Sadly this part is not bisectable. The skiboot side of this change was also sent to the list.

### DIFF
--- a/backends/edk2-compat/edk2-svc-read.c
+++ b/backends/edk2-compat/edk2-svc-read.c
@@ -13,11 +13,9 @@
 #include "external/skiboot/libstb/secvar/secvar.h" // for secvar struct
 #include "backends/edk2-compat/include/edk2-svc.h" // include last, pragma pack(1) issue
 
-static int readFiles(const char *var, const char *file, int hrFlag,
-		     const char *path);
+static int readFiles(const char *var, const char *file, int hrFlag, const char *path);
 static int printReadable(const char *c, size_t size, const char *key);
-static int readFileFromSecVar(const char *path, const char *variable,
-			      int hrFlag);
+static int readFileFromSecVar(const char *path, const char *variable, int hrFlag);
 static int readFileFromPath(const char *path, int hrFlag);
 static int getSizeFromSizeFile(size_t *returnSize, const char *path);
 static int readTS(const char *data, size_t size);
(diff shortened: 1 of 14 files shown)


backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -453,7 +432,7 @@ static int validateSingularESL(size_t *bytesRead, const unsigned char *esl,
 		return ESL_FAIL;
 	}
 	// get certificate
-	cert_size = get_esl_cert((const char *)esl, sigList,
+	cert_size = get_esl_cert((const char *)esl, eslvarsize,
```
nick-child-ibm, line 435: We might want to revisit `get_esl_cert` in skiboot sometime. I remember why I implemented it differently, I did not like how we call `get_esl_signature_list` inside this function despite already having aquired an ESL list object in all of the calling functions. Its a small function, that is more or less a type cast but it irked me.  For not though, it looks good to me.

nick-child-ibm (commented):
builds and passes tests! LGTM!

backends/edk2-compat/edk2-svc-validate.c
daxtens: I'm certainly very happy for you to do that - if you want to propose a patch to skiboot we can rejig this patch to use the updated skiboot version. I will hold off on merging for now.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -412,7 +412,8 @@ static int printReadable(const char *c, size_t size, const char *key)
 		eslsize = sigList->SignatureListSize;
 		printESLInfo(sigList);
 		// puts sig data in cert
-		cert_size = get_esl_cert(c + offset, sigList, (char **)&cert);
+		cert_size =
+			get_esl_cert(c + offset, eslvarsize, (char **)&cert);
```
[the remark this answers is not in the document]
erichte-ibm: Would anyone complain if the column size restriction was either removed or increased? 80 is a bit...narrow.

backends/edk2-compat/edk2-svc-validate.c
nick-child-ibm: I say just merge this. We can add this to the TODO, it's a very small fix that ultimately wouldn't change much. Another thing I would like to add to TODO is change `get_esl_signature_list` to something like `get_efi_signature_list` or `get_efi`, its like saying "ATM machine", the 'M' stands for machine! Anyhows, I am just showing how pedantic I am and that I think these little things can wait until we get bored.

nick-child-ibm:
>     * brings in compiler.h, which is pretty handy

Would you mind explaining what this file does and what makes it _handy_

backends/edk2-compat/edk2-svc-validate.c
erichte-ibm: I may be wrong, but I think that particular function is more of an unpacking of the ESL than actually acquiring one. I think it is grounds for renaming (at some point, not a priority), though I will side channel you a potential concern with naming.

backends/edk2-compat/edk2-svc-read.c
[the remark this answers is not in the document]
daxtens: No complaints from me.

daxtens:
Re compiler.h - it provides short names for `__attribute__((...))` things. For our case it provides `__packed` which we can then use in edk2.h to avoid the pragma pack issues. I can extend the commit message for that change if you like - that commit message just mentions that it is required for edk2.h.

backends/edk2-compat/edk2-svc-read.c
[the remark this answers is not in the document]
daxtens: I've bumped it to 100 in 1 patch, applied all the changes in a separate patch, and this change now goes away.

daxtens:
I've updated the commit message around compiler.h.

nick-child-ibm:
LGTM @erichte-ibm , can we merge?

erichte-ibm:
Looks good to me, merging.

PR #18, opened 2021-05-11
Remove crypto dependency on secvarctl's generic functions

nick-child-ibm:
This pull request edits the `crypto` functions to not depend on secvarctls `generic` functions. Functions defined in `generic.c` are ease-of-use functions that simplify basic I/O tasks. Previously, functions in `crypto` used these `generic` functions to save some effort. With the recent goal to move these `crypto` functions into skiboot, we must remove this dependency. This PR removes all calls from the crypto functions to the generic functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,8 +67,8 @@ endif(  )
 add_executable( secvarctl ${SRC} )
 
 #no crypto means don't compile the generate command = smaller executable
-option( NO_CRYPTO "Build without crypto functions for smaller executable, some functionality lost" OFF )
-if ( NOT NO_CRYPTO )
+option( CRYPTO_READ_ONLY "Build without crypto write functions for smaller executable, some functionality lost" OFF )
+if ( NOT CRYPTO_READ_ONLY )
   target_compile_definitions( secvarctl PRIVATE  SECVAR_CRYPTO_WRITE_FUNC )
 endif(  )
 
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,10 @@ else
 	STATICFLAG=
 endif
 
-#use SECVAR_CRYPTO_WRITE_FUNC for smaller executable but limited functionality
-NO_CRYPTO = 0 
-ifeq ($(strip $(NO_CRYPTO)), 0)
+#use CRYPTO_READ_ONLY for smaller executable but limited functionality
+#removes all write functions (secvarctl generate, pem_to_der etc.)
+CRYPTO_READ_ONLY = 0 
+ifeq ($(strip $(CRYPTO_READ_ONLY)), 0)
 	_CFLAGS+=-DSECVAR_CRYPTO_WRITE_FUNC
 endif
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For any questions regarding secvarctl, feel free to reach out: [Nick Child](nick
  | Build W OpenSSL as cryptolib | `make OPENSSL=1` | `mkdir build && cd build && cmake -DOPENSSL=1 [build options] ../ . && cmake --build .` |
  | Static Build | `STATIC=1` | `-DSTATIC=1`|
  | Reduced Size Build | default | `-DSTRIP=1` |
- | Build Without Crypto Functions | `NO_CRYPTO=1` | `-DNO_CRYPTO=1` |
+ | Build Without Crypto Write Functions | `CRYPTO_READ_ONLY=1` | `-CRYPTO_READ_ONLY=1` |
  | Build W Specific Mbedtls Library | `CFLAGS="-I<path>/include" LDFLAGS="-L<path>/library"` | `-DCUSTOM_MBEDTLS=<path>` |
  | Build for Coverage Tests | `make [options] secvarctl-cov` | `-DCMAKE_BUILD_TYPE=Coverage` |
  | Build W Debug Symbols | `make DEBUG=1` | default |


external/skiboot/libstb/secvar/crypto/crypto-openssl.c
```diff
@@ -667,7 +665,9 @@ int crypto_md_generate_hash(const unsigned char *data, size_t size,
 	if (verbose) {
 		printf("Hash generation successful, %s: ",
 		       OBJ_nid2sn(hashFunct));
-		printHex(*outHash, *outHashSize);
+		for (int i = 0; i < *outHashSize; i++)
+			printf("/%02x", (*outHash)[i]);
```
daxtens, line 669: (I realise that you're copying over your existing behaviour, but...) Wouldn't it be more conventional to use `:` as a separator rather than `/`?

Also, perhaps to avoid the separator preceding the first hash value.
```
printf("%02x", (*outHash)[0]);
for (int i = 1; i < *outHashSize; i++) // (also, does our C spec variant allow loop counters to be declared here?)
    printf(":%02x", ...)
```

external/skiboot/libstb/secvar/crypto/crypto-openssl.c
```diff
@@ -253,39 +253,45 @@ int crypto_pkcs7_generate_w_signature(unsigned char **pkcs7, size_t *pkcs7Size,
...
+			prlog(PR_ERR, "ERROR: failed to open file %s: %s\n",
+			      keyFiles[i], strerror(errno));
 			rc = INVALID_FILE;
 			goto out;
 		}
-		rc = crypto_convert_pem_to_der(
-			crtPEM, crtSizePEM, (unsigned char **)&crt, &crtSize);
-		if (rc) {
+		rc = PEM_read(fp, &unnecessary_name, &unnecessary_hdr, &crt,
+			      (long int *)&crtSize);
```
daxtens, line 285: Hmm, that's a slightly concerning cast. We should probably make crtSize (and keySize) a long and then cast them to the required type - possibly with a test or assert to make sure they're non-negative. Harder to be stuffed around with integer casts than pointer casts, IMO.

external/skiboot/libstb/secvar/crypto/crypto-mbedtls.c
```diff
@@ -14,7 +14,6 @@
 #include "external/extraMbedtls/include/pkcs7.h"
 #include "external/extraMbedtls/include/generate-pkcs7.h"
 #include <mbedtls/platform.h>
-#include "generic.h"
```
daxtens, line 17: Yeah this is a good start. I wonder if we should wait until you have all the parts to get this to compile purely as part of skiboot and then go from there?

daxtens (changes requested):
This is a really good start, I have only minor nits.

I wonder if we should try and bundle together all the changes that are required to get the crypto abstraction layer to compile as part of skiboot, rather than go piecewise into secvarctl? Not wildly fussy though, if you or @erichte-ibm have strong feelings.

nick-child-ibm:
> I wonder if we should try and bundle together all the changes that are required to get the crypto abstraction layer to compile as part of skiboot, rather than go piecewise into secvarctl? Not wildly fussy though, if you or @erichte-ibm have strong feelings.

I think that is a good idea. When I am done getting this crypto stuff to work on SkiBoot, I will just copy it into secvarctl and commit.

external/skiboot/libstb/secvar/crypto/crypto-openssl.c
erichte-ibm: We're not adhering to a particularly restrictive C spec (that I'm aware of), but I would prefer not having initializations inside for-loops.

erichte-ibm:
> > I wonder if we should try and bundle together all the changes that are required to get the crypto abstraction layer to compile as part of skiboot, rather than go piecewise into secvarctl? Not wildly fussy though, if you or @erichte-ibm have strong feelings.
> 
> I think that is a good idea. When I am done getting this crypto stuff to work on SkiBoot, I will just copy it into secvarctl and commit.

Thirding this. I don't think it's a bad idea to keep the crypto abstraction in skiboot, in case we decide to switch to openssl (or some other library) in skiboot for some crazy reason. Ideally if `libstb` is to stand more on its own, then it'd make sense for it to be usable as part of skiboot OR secvarctl without any modification.

external/skiboot/libstb/secvar/crypto/crypto-openssl.c
nick-child-ibm: This may be worthy of an entire new PR. All for loops in secvarctl use internal counter initializations. In terms of C spec, we use gnu99 which is C99 with gnu extensions. It appears that C99 allows for declarations in a for loop. Let me know where we want to go with this.

erichte-ibm: Considering this is a skiboot file, we should adhere to their usage. A quick grep shows at least a few instances of initializing in the loop, but ultimately it will be up to the current maintainer.

I value consistency more than my personal gripes though, so if these are already consistently used in secvarctl, I'm fine with whichever.

nick-child-ibm:
I should have possible made another pull request to redefine its scope but I felt lazy so hopefully this will do:
The new commits were all part of the effort to compile a skiboot branch with the crypto abstraction functions. This has been done [here](https://github.com/nick-child-ibm/skiboot/tree/crypto_devel). Once I got skiboot to compile, with some edits to the `crypto` files, I have now copied those files back into secvarctl so that we are one step closer to submoduling. 

Some interesting things came up in this porting-a-port process. Mainly, we now have 3 versions of both `pkcs7_write.c` and `pkcs7.c`. They both exist in skiboot, mbedtls PR's and secvarctl. I believe, as of this PR, all of the `pkcs7_write.c` are equivalent. But the `pkcs7.c` files all contain different functionalities and return codes. We should look to update these soon. Additionally, while adding `crypto` to skiboot, I realized we don't need any of the `crypto_write...` functions and therefore we do not need `pkcs7_write.c` in skiboot. I added it to the source tree but do not compile it.

erichte-ibm:
>  Additionally, while adding crypto to skiboot, I realized we don't need any of the crypto_write... functions and therefore we do not need pkcs7_write.c in skiboot. I added it to the source tree but do not compile it.

Hmm, the main focus for `NO_CRYPTO` was probably for skiroot (not skiboot). Is there value in still having an actually fully `NO_CRYPTO`'d build? How much functionality is lost? Obviously we can't generate updates, but can we still enroll and display variables?

Either way, we shouldn't need to include `pkcs7_write.c` at all in skiboot, I'd `#ifdef` out any references to those functions in the other secvar/ctl code.

nick-child-ibm:
> Is there value in still having an actually fully `NO_CRYPTO`'d build?
Because the macro name `NO_CRYPTO` implies there being no crypto functions, I have changed it to `CRYPTO_READ_ONLY`.  This is mainly because crypto library functions are still required for reading and updating secvars (things like `mbedtls_x509_parse`). I know this isn't exactly "crypto" but I think it makes more sense as `CRYPTO_READ_ONLY`. At this moment, `CRYPTO_READ_ONLY` completely `#idfef`'s out all of `secvarctl generate` related code as well as all of the `crypto` functions that are specific to `secvarctl generate` . There was one final `#ifdef`ing that occurs is `edk2-svc-validate.c` because it was using a  function in `pkcs7_write.c` (if `CRYPTO_READ_ONLY` then certificates can only be parsed in DER, input certs in PEM are not accepted). I think this makes sense, it has been some time since I examined the difference in executable size but I am not sure what else to remove. 


> Either way, we shouldn't need to include `pkcs7_write.c` at all in skiboot, I'd `#ifdef` out any references to those functions in the other secvar/ctl code.
I have removed pkcs7_write.c and reverted any changes made to `libstb/crypto/pkcs7/` . Rebased and tested everything works fine with preprocessor directives factoring out unnecessary functions.

nick-child-ibm:
Rebased to match `crypto` files with most recent of those in my skiboot branch. The changes are limited to adding a preprocessor directive around `#include pkcs7_write.h` and fixing some whitespace in source code.

nick-child-ibm:
Rebased onto new `main`

CMakeLists.txt
```diff
@@ -67,8 +67,8 @@ endif(  )
 add_executable( secvarctl ${SRC} )
 
 #no crypto means don't compile the generate command = smaller executable
-option( NO_CRYPTO "Build without crypto functions for smaller executable, some functionality lost" OFF )
-if ( NOT NO_CRYPTO )
+option( CRYPTO_READ_ONLY "Build without crypto write functions for smaller executable, some functionality lost" OFF )
+if ( NOT CRYPTO_READ_ONLY )
```
erichte-ibm, line 71: I like this a lot better. :+1: 

Side question: it doesn't matter whether we use either crypto lib for this right?
Side side question: does this work for both static and dynamic linked builds?
Side side side question: can we build static with all crypto libs we support, or is that limited to mbedtls only?

None of the above questions are relevant to this PR, just wondering for future.

external/skiboot/libstb/secvar/crypto/crypto-mbedtls.c
```diff
@@ -1,20 +1,34 @@
...
 
 #include <mbedtls/pk_internal.h> // for validating cert pk data
 #include <mbedtls/error.h>
 #include <mbedtls/md_internal.h>
 #include <mbedtls/oid.h>
+// Nick Child edited paths below
+// #include "libstb/crypto/pkcs7/pkcs7.h"
+// #ifdef SECVAR_CRYPTO_WRITE_FUNC
+// #include "libstb/crypto/pkcs7/pkcs7_write.h"
+// #endif
```
erichte-ibm, line 27: Modification of code in `external`, is this necessary, or is this something we can adjust upstream as well? I know this is a bit of an old PR and I'm not up to date on where things are at the moment.

erichte-ibm:
Mostly looks fine to me, going to defer to @daxtens 's judgement to merge as he still has the open changes requested.

external/skiboot/libstb/secvar/crypto/crypto-mbedtls.c
erichte-ibm: The plan is to eventually have this code identical between skiboot and secvarctl, correct? If that is the case, then these paths should be unified between the two (somehow) so we don't need to modify skiboot code to work with secvarctl.

If there's ongoing effort elsewhere to fix this (possibly what Daniel is working on with libstb-secvar), then we can overlook this for now.

PR #19, opened 2021-05-18
Update tests

nick-child-ibm:
This Pull request allows CI to run successfully with secvarctl build with openssl and mbedtls. Previously, CI was failing due to some tests failing because secvarctl with openssl does not have the same features as secvarctl with mbedtls. This adds a command line flag that jenkins can use for running tests that are crypto lib dependent. Additionally, this commit moves tests that convert auths to esls from the generic python test script to the `secvarctl generate` python test script.

### DIFF
--- a/test/runSvcGenerateTests.py
+++ b/test/runSvcGenerateTests.py
@@ -328,6 +328,10 @@ class Test(unittest.TestCase):
 
 	def test_genExternalSig(self):
 		out = "genExternalSigLog.txt"
+
+		if OPENSSL:
+			command(['echo' , '"TEST NOT RAN, OPENSSL BUILDS DO NOT HAVE THIS FEATURE"' ], out, False)
+			return
 		timestamp = ["-t", "2020-1-1T1:1:1"]
 		inpCrt = "./testdata/db_by_KEK.crt"
 		sigCrt = "./testdata/goldenKeys/KEK/KEK.crt"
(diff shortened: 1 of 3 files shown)


test/runSvcGenerateTests.py
```diff
@@ -328,6 +328,10 @@ def test_genResetFiles(self):
 
 	def test_genExternalSig(self):
 		out = "genExternalSigLog.txt"
+
+		if OPENSSL:
+			command(['echo' , '"TEST NOT RAN, OPENSSL BUILS DO NOT HAVE THIS FEATURE"' ], out, False)
```
daxtens, line 339: it seems mean to pick on spelling in a test case, but builds has a D :P

daxtens (approved):
I'm by no means an expert in the secvarctl test cases, but these certainly look fine to me.

test/runSvcGenerateTests.py
daxtens: (if you choose to fix this, feel free to just fix up the patch and then merge the series without going through another round of reviews!)

nick-child-ibm: Thank you. If you have not noticed yet, my spelling is horrible. I actually really appreciate these comments, its really embarrassing to see silly typos in one of the packaged releases... Software cannot look like it was made by humans, any sign of human error shows weakness! Linus said that once...right?

PR #20, opened 2021-05-19
skiboot/config.h: Remove config file since we have our own now

nick-child-ibm:
With the recent inclusion of a custom skiboot config file, located in `include/config.h` . We no longer have use for `external/skiboot/include/config.h`. This file seems to be lost and actually should be in `external/skiboot/ccan/` but after digging deeper, it seems to be originally used in skiboot only for running tests. It was likely useful to secvarctl prior to having its own `config.h`, that being said there is no use in having both and everything compiles and passes tests without it.


daxtens:
Yeah, you're correct to observe that we don't use this file.

If we submodule or vendor skiboot that will bring a file in this location back (although with different content!), but I don't think that provides a sufficient reason to keep it in the tree now.

Will merge. Thanks for your attention to detail!

PR #21, opened 2021-05-24
pkcs7.c: Respond to bug in sizeof( hash)

nick-child-ibm:
Previously, the pkcs7 code was using sizeof(hash) in pkcs7_verify to store the size of the hash. But, since `hash` is a pointer, the incorrect value (8 instead of 32) was being passed. Mbedtls accepts 0 to use the default value for the hash function (ex: SHA256 == 32 bytes), so this commit makes use of that. Thanks to Daniel Axtens for making this observation.


daxtens:
LGTM, merged.

daxtens:
I think I like the use of 0. It boils down to a call to mbedtls_md_get_size() internally, which is what we would have called anyway. We have to call that earlier to allocate enough memory, but by using 0 here we have a chance to catch any future over-reads with something like valgrind.

PR #22, opened 2021-05-26
edk2-svc-verify.c: Do not add empty secvars to variable bank

nick-child-ibm:
Previously, if an empty (size 0) secvar existed in the sysfs, it was added to the current variable bank (see `setupBanks()`). This leads to errors in `validateBanks` because it calls `validateESL` on all entries in the current variable bank. `validateESL` returns an error if there is no data in the esl. This means that if there is no data in the current KEK, the verify function will return early on an error that the contained data in KEK/data is not a valid ESL. This commit, omits empty secvars from the current list of variables since an empty secvar is a common thing in Secure Boot and should not be treated as an error.

### DIFF
--- a/backends/edk2-compat/edk2-svc-verify.c
+++ b/backends/edk2-compat/edk2-svc-verify.c
@@ -340,8 +340,13 @@ static int setupBanks(struct list_head *variable_bank, struct list_head *update_
 	for (int i = 0; i < currCount; i += 2) {
 		if (defaultVarsFlag) {
 			// if getting secvar successful add tmp to list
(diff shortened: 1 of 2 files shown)


daxtens:
It looks good to me but I have no way of testing it. Is it possible to add a test for this?

nick-child-ibm:
> It looks good to me but I have no way of testing it. Is it possible to add a test for this?

Good point. I have added a test for this in `runTests.py`, if you move this commit before the previous one, you will see that the tests fail and they pass with the new patch in `edk2-svc-verify.c`

Issue #10, opened 2021-03-31
Append/Remove specific entries from secure variables


nasastry:
Current working is any update of secure variable completely replaces the previous value. For example, if the KEK currently contains the ESL's X, Y and Z then using secvarctl  or writing to the sysfs locations to update the KEK with a new auth file containing the ESL's A and B will set the updated KEK to just contain A and B . So, appending and removing specific entries is currently not supported.
This is the request for feature/enhancement to support append/remove of specific entries from the secure variables.
Thank you.

nick-child-ibm:
Hey Nageswara, thanks for your patience! I finally got around to implementing half of your request. I will get around to `secvarctl remove` soon

nick-child-ibm:
Sorry, forgot to update this. Pull request #23 should have fully implemented the requested features (`insert` and `remove` and test cases). Just waiting on review.

PR #23, opened 2021-05-28
Add new insert and remove subcommand

nick-child-ibm:
In response to [Nageswara's recommendation](https://github.com/open-power/secvarctl/issues/10), this pull request adds a new secvarctl subcommand: insert. `secvarctl insert` allows for a new secvar update to be submitted without the old one being replaced. The use case is if the user would like to add an ESL without overwriting the previous secvar ESLs.  

This PR comes complete with implementations, tests and documentation for the new `secvarctl insert` command. 

Okay, so I sorta-kinda lied when I said "`secvarctl insert` allows for a new secvar update to be submitted without the old one being replaced" . In reality, the function gets the current ESLs and appends the new ESL to it. With the combined ESL, it is sent to be put into an auth file. With the auth file, `secvarctl insert` allows the user to use the `-w` flag to submit the update to secvar sysfs or to a file with `-o <outFile>`. So, to simplify, the function really takes in two ESL's-> appends them -> generate auth -> ouptut . Various flags were added for different use cases. For example, by default the current ESL is taken from secvar sysfs `/sys/firmware/secvar/vars/<varName>/data` but the user can use `-e <eslFile>` to specify the current ESL. Additionally, the `-p PATH` flag can be used to specify the sysfs where it writes (to `PATH/<varName>/update`) and reads (to `PATH/<varName>/data`

Special thanks to @nasastry for a great feature request! I promise `secvarctl remove` is coming soon...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ set( CMAKE_C_COMPILER gcc )
 set( SRC secvarctl.c generic.c )
 
 #sources for edk2 backend
-set ( EDK2SRC edk2-svc-read.c edk2-svc-write.c edk2-svc-validate.c edk2-svc-verify.c edk2-svc-generate.c )
+set ( EDK2SRC edk2-svc-read.c edk2-svc-write.c edk2-svc-validate.c edk2-svc-verify.c edk2-svc-generate.c edk2-svc-insert.c )
 set ( EDK2SRCDIR backends/edk2-compat/ )
 list( TRANSFORM EDK2SRC PREPEND ${EDK2SRCDIR} )
 list( APPEND SRC ${EDK2SRC} )
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ _LDFLAGS += -s
 endif
 
 EDK2OBJDIR = backends/edk2-compat
-_EDK2_OBJ =  edk2-svc-read.o edk2-svc-write.o edk2-svc-validate.o edk2-svc-verify.o edk2-svc-generate.o
+_EDK2_OBJ =  edk2-svc-read.o edk2-svc-write.o edk2-svc-validate.o edk2-svc-verify.o edk2-svc-generate.o edk2-svc-insert.o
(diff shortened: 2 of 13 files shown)


nick-child-ibm:
`secvarctl insert` implemented and has  test suite. `secvarctl remove` implemented but holding off on writing tests until I get the go ahead on general approach and arrguments (using X509 serial number as ESL identifier). Also still have to implement insert and remove for `dbx` updates.

nick-child-ibm:
Okay, I have finally got back to this. Thanks for all your patience. Updates are as follows:
1. rebased on `main` and fix conflicts
2. added test cases for `secvarctl remove`
3. implemented `insert/remove` for the dbx (`insert` works the same as the other variables, `remove` not so much)
4. add test cases for `secvarctl insert/remove` when working with dbx
5. update documentation

The way `remove` is currently working is to use the `-x <serial_number>` flag to identify an ESL that is to be removed from a chain of ESL's. The `<serial_number>` must be in the format  `XX:XX:XX...:X` with 20 bytes of hex data and be contained in the X509 of an ESL. This value can be taken from reading and ESL with  `secarctl read` or `secvarctl -v -e <esl_file>`. I could not think of an easier way to identify the ESL to remove from the chain, besides indexing them but that feels like a recipe for index-at-0 vs index-at-1 user issues. Anways, I had figured the serial number would be sufficient BUT the RFC says that it technically does not need to be 20 bytes (it just has to have a max of 20 bytes) AND it is only a unique value between it and it's CA's. But who is really going to generate there own X509's (openssl generates with 20 bytes and does not allow for custom values) ? right? right? Looking for more thoughts on this. Also, since the dbx does not contain x509's, the `-x <serial_number>` is interpreted as `-x <hash_value>` and uses the same `XX:XX...X` format to identify a hash contained in an ESL of the dbx's ESL chain. 

Looking forward to more feedback!
Thanks!
Nick

erichte-ibm:
Stale, and no longer a focus target. Closing PR.

PR #24, opened 2021-06-16
Do not install man pages into /usr/local

hramrach:
the binary is installed into /usr/bin and the man page into
/usr/local/man. That does not make any sense.

Signed-off-by: Michal Suchanek <msuchanek@suse.de>


nick-child-ibm:
@hramrach  Thanks for the pull request. Good find!

Issue #25, opened 2021-06-22
secvarctl read/write throws warning and error


nasastry:
secvarctl tool throws the following warning and error while read/write operations.

```
ERROR: failed to read whole contents of /sys/firmware/secvar/format in one go
WARNING!! Could not extract data from /sys/firmware/secvar/format , assuming platform does not support secure variables
WARNING: Unsupported backend detected, assuming ibm,edk2-compat-v1 backend
Read/write may not work as expected
```

Size and content of the sysfs file `/sys/firmware/secvar/format `

```
# cat /sys/firmware/secvar/format
ibm,edk2-compat-v1
# ls -l /sys/firmware/secvar/format
-r--r--r-- 1 root root 65536 Jun 22 09:37 /sys/firmware/secvar/format
```

PR #26, opened 2021-06-29
add max_len feature to getDataFromFile

nick-child-ibm:
This PR was in response to #25 . The issue was that we were trying to read all of a large file, when we only needed the first couple bytes. This issue is actually not unusual when reading secure variable files. In fact, when reading secvars we only read the amount of data specified in `<secvar>/size` since `<secvar>/data` is usually a much larger file than the data it contains. Since file reading processes have already been implemented in a couple different places and patterns are starting to grow, I decided to edit the function `getDataFromFile` (we really should switch to snake case sometime soon) to accept a parameter `max_len` so it only reads a max number of bytes. If the calling function wishes to read the entire file, they can use `max_len = 0`. This allowed me to go back to `getSecvar` and use the updated `getDataFromFile` call instead of re-implementing file read logic. 

Thanks to @nasastry for finding this issue.

### DIFF
--- a/secvarctl.c
+++ b/secvarctl.c
@@ -112,14 +112,21 @@ int main(int argc, char *argv[])
 static struct backend *getBackend()
 {
 	char *buff = NULL, *secVarFormatLocation = "/sys/firmware/secvar/format";
-	size_t buffSize;
+	size_t buffSize, max_buff_size;
 	struct backend *result = NULL;
+	int i;
 	// if file doesnt exist then print warning and keep going
 	if (isFile(secVarFormatLocation)) {
 		prlog(PR_WARNING, "WARNING!! Platform does not support secure variables\n");
 		goto out;
 	}
-	buff = getDataFromFile(secVarFormatLocation, &buffSize);
+	// get max size of backend name
+	max_buff_size = strlen(backends[0].name);
+	for (i = 0; i < sizeof(backends) / sizeof(struct backend); i++)
+		if (strlen(backends[i].name) > max_buff_size)
+			max_buff_size = strlen(backends[i].name);
+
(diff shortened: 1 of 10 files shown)


nasastry:
@nick-child-ibm Tested the above patch and not seeing any `warnings or errors` as reported earlier. Thanks for the patch.

secvarctl.c
```diff
@@ -112,22 +112,29 @@ int main(int argc, char *argv[])
...
 	struct backend *result = NULL;
+	int i;
 	// if file doesnt exist then print warning and keep going
 	if (isFile(secVarFormatLocation)) {
 		prlog(PR_WARNING, "WARNING!! Platform does not support secure variables\n");
 		goto out;
 	}
-	buff = getDataFromFile(secVarFormatLocation, &buffSize);
+	// get max size of backend name
+	max_buff_size = strlen(backends[0].name);
```
erichte-ibm, line 124: Perhaps use a constant/macro for this in the header (or wherever `backends[]` is defined), or something else that can be done at compile-time. This is going to have to be updated every time a new backend is added, with a potentially longer name.

Personally, I'd suggest defining a reasonable maximum in a macro constant next to the table, maybe ~64 bytes or so. If we end up with a longer backend name, that number can be incremented at the same time.

erichte-ibm: The alternative is to also loop through the backends first to get the maximum format name length, would also prevent needing to ever adjust this code in the future. Bit excessive though.

erichte-ibm: Ah that's what I get for speedreading.

Hmmm... this is fine then, ignore what I said. There's probably a way to do this fully in preprocessor logic, but that's a needless optimization.

PR #28, opened 2021-07-27
BUG: edk2-compat.c: Turn on setup mode if PK is updated

nick-child-ibm:
Previously, we are commmenting out the section of code that enforced secure boot if the the PK is updated. I likely thought it was HW key related so I commented it out. I uncommented so now secure boot will be enforced as soon as a PK is enrolled.

This needs a test case but is a bug so opening the PR now.

### DIFF
--- a/external/skiboot/libstb/secvar/backend/edk2-compat.c
+++ b/external/skiboot/libstb/secvar/backend/edk2-compat.c
@@ -207,14 +207,14 @@ static int edk2_compat_process(struct list_head *variable_bank,
 			 * hw-key-hash of that firmware. When PK is updated, hw-key-hash
 			 * is updated. And when PK is deleted, delete hw-key-hash as well
 			 */
-			/*if(neweslsize == 0) { //NICK REMOVED BECAUSE NO HW KEYS
+			if(neweslsize == 0) {
 				setup_mode = true;
-				delete_hw_key_hash(&staging_bank);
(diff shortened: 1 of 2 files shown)


nick-child-ibm:
Edited commit to include a test!

external/skiboot/libstb/secvar/backend/edk2-compat.c
```diff
@@ -207,14 +207,14 @@ static int edk2_compat_process(struct list_head *variable_bank,
...
 			 */
-			/*if(neweslsize == 0) { //NICK REMOVED BECAUSE NO HW KEYS
+			if(neweslsize == 0) {
 				setup_mode = true;
-				delete_hw_key_hash(&staging_bank);
+				// delete_hw_key_hash(&staging_bank); //NICK REMOVED BECAUSE NO HW KEYS
 			} else  {
 				setup_mode = false;
-				add_hw_key_hash(&staging_bank);
+				// add_hw_key_hash(&staging_bank);
```
erichte-ibm, line 215: How many more of these changes are needed still to properly vendor the skiboot code here? Can we possibly just stub out these functions instead?

erichte-ibm:
Left a comment regarding stubbing functions, but merging anyway as this is relatively minor. Stub functions can come when we fully vendor in the skiboot code.

PR #29, opened 2021-08-10
Allow secvarctl to compile with mbedtls V3

nick-child-ibm:
Recently, Mbedtls came out with version[ 3.0.0](https://github.com/ARMmbed/mbedtls/releases/tag/v3.0.0) . The relevant changes include turning all struct fields to private, requiring random number generators for most pk functions and moving around/deleting some files. 

Due to API changes in the two mbedtls versions (2.x and 3.x), I propose that we support both. The user can set the mbedtls version to expect when compiling with the flag `MBEDTLS_V3=1` . There are MBEDTLS macros that we could use to figure out the mbedtls version internally (`MBEDTLS_VERSION_MAJOR` in particular) but it seems more beneficial to keep the two versions separated in two different files and only compile the files we will need. Otherwise preprocessor directives could get real sloppy in `crypto-mbedtls.c` and `pkcs7.c`. I suppose we could separate the files to `v2` and `v3`, ifdef the whole file and compile both versions. I am open to opinions on this. What really would be nice is if we could internally figure out the mbedtls version in the make/cmake files.

For now, there are two versions of `extraMbedtls` (containing `pkcs7` and `pkcs7_write`) and two versions of `crypto-mbedtls`. Depending on the make/cmake options, only one version is compiled. Defaulting to mbedtls v2.


nick-child-ibm:
Note to future Nick, if this gets merged: Remember to update the Fedora spec file for passing the correct mbedtls version flag to Cmake!

external/skiboot/libstb/secvar/crypto/crypto-mbedtls-v3.c
```diff
@@ -0,0 +1,240 @@
+// SPDX-License-Identifier: Apache-2.0
+/* Copyright 2021 IBM Corp.*/
+#ifdef MBEDTLS_V3
+// ^extra precaution to not compile with mbedtls version 3.x unless specified
```
erichte-ibm, line 4: This isn't necessary, and the compilation should fail if somehow both 2.x and 3.x are selected, OR compiling with 3.x is attempted without actually having 3.x

nick-child-ibm:
> If the pkcs7 work goes upstream to mbedtls will that only be for 3.x, or is there a chance that will be backported to 2.x? 
Doesn't look like it ->  [we are being pushed until at least q4](https://github.com/ARMmbed/mbedtls/pull/3431#issuecomment-897417442)
> Whenever that does land upstream, I'm thinking we might want to just enforce that as a minimum version.
That makes sense to me! But until then should we support both mbedtls versions?

erichte-ibm:
> >Whenever that does land upstream, I'm thinking we might want to just enforce that as a minimum version.
> That makes sense to me! But until then should we support both mbedtls versions?
For now we should support both, but we should enforce whatever version supplies pkcs7 as minimum when that is possible. Perhaps we'll bump secvarctl's version to 1.0 with that.

nick-child-ibm:
This branch has gone stale. Commit [28cdd62](https://github.com/open-power/secvarctl/pull/29/commits/28cdd624fcafa3cf3de51c9a05e2f27648536822) copied the pkcs7 files into v2 and v3. The pkcs7 files upstream have since been edited. So the copy in commit [28cdd62](https://github.com/open-power/secvarctl/pull/29/commits/28cdd624fcafa3cf3de51c9a05e2f27648536822) is a copy of old functions. Therefore this work should really be redone entirely. Not to mention there are several merge commits and this branch was created before the unification of error codes was introduced. I am closing this pull request as the work should be redone on the current `main` branch.

PR #30, opened 2021-08-18
 Fix openssl return code in crypto_pkcs7_signed_hash_verify

nick-child-ibm:
Since crypto_strerror is called on the return code of
crypto_pkcs7_signed_hash_verify, we need to make sure that
an openssl error code is returned.

To get a return code, the openssl function ERR_get_error is called.
I am considering making this call in the rest of the file. The only concern
is that it is not uncommon for these openssl functions to decide to NOT set
an error code before returning NULL or 0 or whatever generic fail they return.
In fact it seems only a select few actually add a error code to the error stack
before returning. I am curious to hear more thoughts on this.

We should probably decide if we want to return the openssl/gnutls/mbedtls
error codes in all the functions, or just this specific one. It seems possible/easy
with gnutls and mbedtls but openssl likes to return 1 on success and return pointers
to data instead of return values. Let me know


nick-child-ibm:
Okay closing this to open a new pull request with greater scope.

PR #31, opened 2021-08-25
Allow for static builds to work again

nick-child-ibm:
Not sure when or why but static builds have stopped working in both Cmake and Make builds. `Makefile` needed some more linked libraries and rearrangement of the linking flags. `CMakeLists.txt` was matching with the shared libraries (`.so`) before matching with the `.a` libs. I believe I got everything working with mbedtls builds but openssl is segfaulting at runtime when building statically. It could be an issue with my specific `libopenssl.a` but I will look into it later on. For now, we need to allow for Cmake static mbedtls builds to run smoothly in order to continue with op-build work.


erichte-ibm:
Closing since we likely won't need this anymore. Can re-open if the situation changes, but will be easier for maintenance to drop static builds for now.

PR #35, opened 2021-11-17
Add github actions for CI testing and building rpm's

nick-child-ibm:
Hello,

This pull request adds two new workflows to secvarctl:
1. run CI tests on every push and pull request
2. build x86 rpm and source rpm on every release

Some notes on these:
- CI tests run  on an Ubuntu x86 github runner, it does not seem that github supports actions on other architectures
- I initially thought I could cross compile a ppc64le rpm, it builds but when I try to install on a ppc64le machine I fail to install
- Since ppc64le rpm is failing, for now I commented that part of the workflow out and figured the source rpm would be sufficient

### DIFF
--- a/.github/workflows/build_rpm.yml
+++ b/.github/workflows/build_rpm.yml
@@ -1,0 +1,164 @@
+name: Build RPMs on a new release
+
+on:
+  release:
+  workflow_dispatch:
+  
+# env:
+#     repo_url: ${{github.server_url}}/${{github.repository}}/archive/${{github.ref}}.tar.gz
+#     version: ${{ github.event.release.tag_name }}
+#     archive_file: ${version}.tar.gz
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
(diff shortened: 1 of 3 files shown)


.github/workflows/build_rpm.yml
```diff
@@ -0,0 +1,165 @@
...
+      - name: show spec file
+        run: cat secvarctl.spec
+      - name: setup buildroot
+      #rpmbuild expects specfile and v<specfile version>.tar.gz
+        run: |
+             mkdir ~/rpmbuild
+             mkdir ~/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
+             cp secvarctl.spec ~/rpmbuild/SPECS/
+             cp ${archive_file} ~/rpmbuild/SOURCES/v`sed -n 's/.*Version:\s*\([^ ]*\).*/\1/p' secvarctl.spec`.tar.gz
+       # install package dpeendencies
```
erichte-ibm, line 84: .

Issue #37, opened 2022-09-20
compilation fails with openssl3


nasastry:
Tried to compile on the system where openssl-3.0.1-41 installed and seeing the following 

```
# make OPENSSL=1
make -C ../ secvarctl-cov
make[1]: Entering directory '/home/secvarctl-main'
gcc   -MMD -O2 -std=gnu99 -I./ -Iinclude/ -Iexternal/skiboot/ -Iexternal/skiboot/include -Wall -Werror -DSECVAR_CRYPTO_WRITE_FUNC -DSECVAR_CRYPTO_OPENSSL -c  --coverage external/skiboot/libstb/secvar/crypto/crypto-openssl.c -o external/skiboot/libstb/secvar/crypto/crypto-openssl.cov.o
external/skiboot/libstb/secvar/crypto/crypto-openssl.c: In function 'crypto_x509_get_pk_bit_len':
external/skiboot/libstb/secvar/crypto/crypto-openssl.c:518:9: error: 'EVP_PKEY_get1_RSA' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  518 |         rsa = EVP_PKEY_get1_RSA(pub);
      |         ^~~
In file included from /usr/include/openssl/x509.h:29,
                 from external/skiboot/libstb/secvar/crypto/crypto.h:10,
                 from external/skiboot/libstb/secvar/crypto/crypto-openssl.c:8:
/usr/include/openssl/evp.h:1348:16: note: declared here
 1348 | struct rsa_st *EVP_PKEY_get1_RSA(EVP_PKEY *pkey);
      |                ^~~~~~~~~~~~~~~~~
external/skiboot/libstb/secvar/crypto/crypto-openssl.c:524:9: error: 'RSA_bits' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  524 |         length = RSA_bits(rsa);
      |         ^~~~~~
In file included from /usr/include/openssl/x509.h:36,
                 from external/skiboot/libstb/secvar/crypto/crypto.h:10,
                 from external/skiboot/libstb/secvar/crypto/crypto-openssl.c:8:
/usr/include/openssl/rsa.h:203:27: note: declared here
  203 | OSSL_DEPRECATEDIN_3_0 int RSA_bits(const RSA *rsa);
      |                           ^~~~~~~~
[...]

PR #38, opened 2022-09-26
Build with openssl v3

nick-child-ibm:
Current builds fail with openssl library versions >= 3.0

See #37


nick-child-ibm:
This is currently worked around with a compile time check. This is different from how we are handling mbedtls 2.x vs mbedtls 3.x (see #29 ) . I chose to do this because the differences in the API is actually not as difficult as the mbedtls upgrade.

PR #39, opened 2022-11-07
secvarctl: Support sbat on generate/read/validate

SudhakarKuppusamy1:
Currently, secvarctl supports PK, KEK, db and dbx variable on generate/read/validate. 
However, it is not allowing sbat variable on the same.
So, adding sbat variable supports on generate/read/validate.

SBAT Auth file generation:

    ./secvarctl generate f:a -k <signerPrivate.key> -c <signerPublic.crt> -i <sbatFile> -o <out.auth> -n sbat

SBAT ESL file generation:

    ./secvarctl generate f:e -i <sbatFile> -o <out.esl> -n sbat

SBAT Auth file read:

    ./secvarctl read -f <authFile> -r

SBAT ESL file read:

    ./secvarctl read -f <ESLFile>  sbat

SBAT Auth file validation:

    ./secvarctl validate -a <authFile> -n sbat

SBAT ESL file validation:

    ./secvarctl validate -e <ESLFile> -n sbat

Signed-off-by: SudhakarKuppusamy <sudhakar@linux.ibm.com>


nick-child-ibm:
Thanks for the pull request.
I have not reviewed the entire commit but I am noticing that the test cases are failing. `cd test; make`. This is expected because there is no mock data for the SBAT variable in our mock sysfs (see `./testdata/goldenKeys/`).

But before going and changing the test environment/scripts, I think we should discuss how this patch should effect functionality for host secure boot systems. As we know, host secure boot does not have sbat (and on the other hand guest secure boot does not have a TS IIRC).

Unfortunately, this is the first patch that introduces differences in expected behavior between host secure boot and guest secure boot. So we need to decide, should secvarctl:
 - compile differently for host vs guest secure boot
 - should the advertised secure boot variables be a superset of all known variables or be subset depending on if working with host or guest secure boot
 - should we introduce a new command line flag for specifying the expected backend
 - how should we handle this in our test scripts (do we need a new script for testing on guest environment)

The answer to these questions will determine how we implement SBAT support. Some of these points may be out of scope for SBAT support but still worth considering as soon as possible. I would like some feedback from @erichte-ibm  to weigh in on what should be considered now and what we can implement later on. In my opinion the sooner we prepare for different backends, the easier it will be to maintain further differences between host/guest secure boot. 
As discussed in the latest secvarctl meeting, I believe we decided, as a long-term goal, to share as much code as possible (perphaps by having shared logic in a new directory `backends/common`) but ultimately factoring out the differences in `backends/edk2-compat` and `backends/guest-sb` or something along those lines.

Again, doing all this work is tedious, but getting some amount of it done now will make the addition of SBAT support and further backend differences seamless down the road. 
 
Handling one thing at a time here, I would like to first tackle how `secvarctl read` should work. At the moment `secvarctl read` will go and try to read `/sys/firmware/secvar/vars/[PK, KEK, db ,dbx, sbat, TS]` regardless of host or guest backends. This is an issue because on host it will fail since sbat is nonexistent and on guest it will fail because TS (or dbx, I don't know) is nonexistent. This is a good place to start figuring out:
 1. What is is the expected backend environment (can we figure out intuitvely at run time or do we need a flag or both)?
 2. What are the expected variables for that backend?
 
Apologies for the word dump here and I know it is confusing, I will like to open it up for discussion now. It may be more straightforward to do pull requests for sbat support one secvarctl subcommand at a time.

SudhakarKuppusamy1:
@nick-child-ibm 
Sorry I forgot to add mock data for sbat. I understood the issue and Thank you for the detailed information. 
I have modified the **validate** functionality  **"-x"**  to  **"-n"** (**which is name of secure boot variable, used when validating dbx/sbat**) since dbx contain a hash  and  sbat contain a string but both not an x509.So, it leads to the validate functionality failure in test script.

erichte-ibm:
My first inclination is to merge this as a separate branch for now, so it can be used immediately. Since this is primarily for guest secure boot though, I think the full implementation should be part of the forthcoming `libstb-secvar` backend.

Alternatively, as long as this doesn't cause trouble for host secure boot, we can merge this temporarily, then revert it when the full `libstb-secvar` backend is merged. I will think on this a bit more as I review the code.

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -119,8 +119,8 @@ static int parse_opt(int key, char *arg, struct argp_state *state)
 		verbose = PR_DEBUG;
 		break;
 	// set varname as dbx, important for validating ESL
-	case 'x':
-		args->varName = "dbx";
+	case 'n':
+		args->varName = arg;
```
erichte-ibm, line 123: If I'm reading this correctly, this also changed the generate command for dbx now, right?
So sbat is:
```
./secvarctl generate f:a -k <signerPrivate.key> -c <signerPublic.crt> -i <sbatFile> -o <out.auth> -n sbat
```
and now dbx will be
```
./secvarctl generate f:a <args> -n dbx
```
correct?

This is a small change in the argument usage that would be breaking for any scripts, but I do like the more unified approach over having a specialized argument for only `dbx`.

@nick-child-ibm any thoughts on this?

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -429,6 +431,8 @@ static int validateSingularESL(size_t *bytesRead, const unsigned char *esl, size
 			prlog(PR_INFO, "\tHash: ");
 			printHex(cert, cert_size);
 		}
+	}else if (varName && !strcmp(varName, "sbat")) {
+			rc = SUCCESS;
```
nick-child-ibm, line 435: We need to discuss what is contained in an sbat and if we should be checking its contents

nick-child-ibm, line 435: Since this is getting merged as a seperate branch, we can let the formatting go unchecked, but please add a print statement if `verbose` is specified, similar how the dbx is handled right above it.

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -332,13 +334,42 @@ int validateESL(const unsigned char *eslBuf, size_t buflen, const char *key)
...
+			free(data);
+			return 0;
+		}
+	}
+
+	free(data);
+
+	return 1;
+}
+
```
nick-child-ibm, line 365: Looks like this function is a source of a few buffer overreads. This is not a major concern and is due to `strtok_r` and `strdup` reading past the length of `sbat_data` since `\0` is not found in the sbat buffer. It is encouraged that we address this just to ensure that valgrind is happy and future tests can pass.

Just a suggestion:
```suggestion
static int validate_sbat (const uint8_t *sbat_data, size_t sbat_len)
{
	int number_of_commas = 0;

	for (size_t i = 0; i < sbat_len; i++) {
		if (sbat_data[i] == ',')
			number_of_commas++;
		if (sbat_data[i] == '\n') {
			if (number_of_commas != 1)
				return 0;
			number_of_commas = 0;
		}	
	}

	return 1;
}
```

backends/edk2-compat/edk2-svc-validate.c
```diff
@@ -429,6 +460,17 @@ static int validateSingularESL(size_t *bytesRead, const unsigned char *esl, size
...
+	}else if (varName && !strcmp(varName, "sbat")) {
+		if (!validate_sbat(cert)) {
+			prlog(PR_ERR,
+			      "ERROR: sbat data format is invalid\n");
+			rc = SBAT_FAIL;
+		} else
+			rc = SUCCESS;
+
+		if (verbose >= PR_INFO) {
+			prlog(PR_INFO, "\tSBAT: \n%s\n", cert);
```
nick-child-ibm, line 472: This is another cause of failed memcheck tests. To fix use `printRaw`.

backends/edk2-compat/edk2-svc-read.c
```diff
@@ -374,6 +374,8 @@ static int printReadable(const char *c, size_t size, const char *key)
 		if (key && !strcmp(key, "dbx")) {
 			printf("\tHash: ");
 			printHex(cert, cert_size);
+		} else if (key && !strcmp(key, "sbat")) {
+			printf("\tFound sbat Info:\n%s\n", cert);
```
nick-child-ibm, line 378: `cert` is not null terminated. So valgrind reports buffer over-read. To fix, use secvarctl function `printRaw`.

nick-child-ibm (commented):
Looks like formatting checks are failing (not a big deal, again this is a separate branch so not a big deal). If you want to fix it should be as easy as `make format`.

 But bigger issue is valgrind memory checks are failing. I debugged a bit and left comments on workarounds.
 
 Thanks again for the contribution.

backends/edk2-compat/edk2-svc-read.c
SudhakarKuppusamy1: Fixed it

backends/edk2-compat/edk2-svc-validate.c
SudhakarKuppusamy1: fixed it

SudhakarKuppusamy1:
@nick-child-ibm , Addressed the review comments

nick-child-ibm:
Thanks for your continued work on this @SudhakarKuppusamy1 . Looks good to me, as discussed, merging into new branch `guest-sb-support`

PR #45, opened 2023-08-01
Replace valgrind with ASAN/UBSAN, overhaul debug build

erichte-ibm:
Depends on #44 

Valgrind is a great tool, but has a lot of runtime overhead -- especially at startup. We make probably hundreds of calls to `secvarctl`, and therefore invoke valgrind hundreds of times, causing a large amount of time waiting for valgrind to initialize.

This PR adds new support to the `Makefile` for building with address and undefined behavior sanitizers, which introduce a negligible amount of overhead to running the binary, with the benefit of catching even more problems than valgrind.

This comes with a free rework to how testing/debuild builds are handled:
 - `secvarctl-cov` has been renamed to `secvarctl-dbg` (and corresponding build targets)
 - `make debug` is aliased to building the debug binary
 - all `.cov.o` objects are now `.dbg.o`
 - debug build now enables coverage, ASAN/UBSAN by default
 - ASAN/UBSAN can be disabled with `DISABLE_ASAN=1`
 - debug and release flags have been defined separately:
   - debug flags now enable `-g -O0` by default
   - release flags now use `-s -O2` by default
 - in a similar vein `-s` (strip) is no longer enabled on all builds, making debugging stack traces impossible
 - `$(INCLUDES)` is now its own variable and not injected into `$(CFLAGS)` for some reason

NOTE: for the sake of scope, the test runners have been left alone, as those scripts are probably next on my list to clean up a bit. The `MEMCHECK` option for running with valgrind will be removed in a later PR.

TL;DR - all debug builds include coverage support and memcheck'd by default, and tests now run like 20x faster or something.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022-2023 IBM Corp.
-CC = gcc
-_CFLAGS = -MMD -O2 -std=gnu99 -Wall -Werror
-CFLAGS =
+CC ?= gcc
+_CFLAGS = -MMD -std=gnu99 -Wall -Werror
 # TODO: just put all the linker flags for now, rework the LDFLAGS settings later
 LDFLAGS = -lcrypto -lmbedtls -lmbedx509 -lmbedcrypto
 MANDIR=usr/share/man
@@ -15,13 +14,6 @@ INCLUDES = -I.                                       \
            -I./external/libstb-secvar/               \
            -I./external/libstb-secvar/include        \
            -I./external/libstb-secvar/include/secvar
-
-DEBUG ?= 0
-ifeq ($(strip $(DEBUG)), 1)
-  _CFLAGS += -g
-else
-  _LDFLAGS += -s
-endif
 
 #use CRYPTO_READ_ONLY for smaller executable but limited functionality
 #removes all write functions (secvarctl generate, pem_to_der etc.)
@@ -130,6 +122,9 @@ SANITIZE_FLAGS = -fsanitize=address              \
                  -fno-sanitize=null              \
                  -fno-sanitize=alignment
 
+DEBUG_CFLAGS = -g -O0 --coverage
+RELEASE_CFLAGS = -s -O2
+
 export CFLAGS
 export LDFLAGS
 
@@ -139,11 +134,11 @@ all: secvarctl
 
 $(OBJ_DIR)/%.o: %.c
 	@mkdir -p $(dir $@)
-	$(CC) $(_CFLAGS) $< -o $@ -c
+	$(CC) $(RELEASE_CFLAGS) $(_CFLAGS) $(INCLUDES) $< -o $@ -c
 
 $(OBJ_DIR)/%.dbg.o: %.c
 	@mkdir -p $(dir $@)
-	$(CC) $(_CFLAGS) -c --coverage -g $(SANITIZE_FLAGS) $< -o $@
+	$(CC) $(DEBUG_CFLAGS) $(_CFLAGS) $(SANITIZE_FLAGS) -c  $< -o $@
 
 $(BIN_DIR)/secvarctl: $(OBJS) $(LIBSTB_SECVAR)
 	@mkdir -p $(BIN_DIR)
@@ -151,7 +146,7 @@ $(BIN_DIR)/secvarctl: $(OBJS) $(LIBSTB_SECVAR)
 
 $(BIN_DIR)/secvarctl-dbg: $(OBJDBG) $(LIBSTB_SECVAR)
 	@mkdir -p $(BIN_DIR)
-	$(CC) $(_CFLAGS) $^ $(STATICFLAG) -g $(SANITIZE_FLAGS) -fprofile-arcs -ftest-coverage -o $@ $(_LDFLAGS)
+	$(CC) $(_CFLAGS) -g $^ $(STATICFLAG) $(SANITIZE_FLAGS) -fprofile-arcs -ftest-coverage -o $@ $(_LDFLAGS)
 
 $(LIBSTB_SECVAR):
 	$(MAKE) CFLAGS=-DSECVAR_CRYPTO_WRITE_FUNC -C external/libstb-secvar lib/libstb-secvar-openssl.a
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ DEPS = $(OBJS:.o=.d)
 _CFLAGS += $(CFLAGS) $(INCLUDES)
 _LDFLAGS += $(LDFLAGS)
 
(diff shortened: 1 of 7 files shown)


test/Makefile
```diff
@@ -17,25 +16,21 @@ GUEST_BACKEND = 1
...
-	$(py) runTests.py -m $(1) -s $(SECVAR_TOOL) && \
-	$(py) runSvcGenerateTests.py -m $(1) -o $(OPENSSL) -g $(GNUTLS) -s $(SECVAR_TOOL)
+	$(py) runTests.py -s $(SECVAR_TOOL) && \
+	$(py) runSvcGenerateTests.py -o $(OPENSSL) -g $(GNUTLS) -s $(SECVAR_TOOL)
 endef
 
 define test_guest
 	@cd guest/ && \
-	$(py) guest_tests.py -m $(1) -s $(SECVAR_TOOL)
+	$(py) guest_tests.py -s $(SECVAR_TOOL)
```
nick-child-ibm, line 25: please remove the `-m` arguments from guest and host python tests. Additionally any `MEMCHECK` and`valgrind` references can be removed as well.

Makefile
```diff
@@ -134,39 +141,33 @@ $(OBJ_DIR)/%.o: %.c
...
 
 secvarctl: $(BIN_DIR)/secvarctl
-secvarctl-cov: $(BIN_DIR)/secvarctl-cov
-.PHONY: secvarctl secvarctl-cov
+secvarctl-dbg: $(BIN_DIR)/secvarctl-dbg
+debug: $(BIN_DIR)/secvarctl-dbg
+.PHONY: secvarctl secvarctl-dbg debug
 
-check: secvarctl-cov
+check: secvarctl-dbg
```
nick-child-ibm, line 162: we are going to need to update documentation at some point. README and man file will probably need a new review.

nick-child-ibm (approved):
`make check` is failing for me:
```
$ secvarctl-stuff/erichte-guest-devel/secvarctl/bin/secvarctl-dbg -m host generate -v reset -n db -k ./testdata/goldenKeys/KEK/KEK.key -c ./testdata/goldenKeys/KEK/KEK.crt -o ./generatedTestData/reset_db_by_KEK.auth

WARNING!! platform does not support secure variables
WARNING: unsupported backend detected, assuming ibm,edk2-compat-v1
read/write may not work as expected
Input file is empty of type reset , output file is ./generatedTestData/reset_db_by_KEK.auth of type auth
Timestamp is : 2023-08-04 16:28:40 UTC
Combining Auth header, PKCS7 and new ESL:
	+ Auth Header 40 bytes
	+ PKCS7 1236 bytes
backends/host/host_svc_generate.c:1111:2: runtime error: null pointer passed as argument 2, which is declared to never be null
```

I believe we are safe since this step is appending an empty ESL onto the auth file. So the faulty line:
`memcpy(*outBuff + offset, newESL, eslSize);`
translates to:
`memcpy(ptr, NULL, 0);`
Which is actually by design:
```
 case 'r':
                // if creating a reset key, ensure input is NULL and size of zero
                if (inpSize == 0 && *inpPtr == NULL)
                        rc = SUCCESS;
```
this shuts up the error though:
```
-       memcpy(*outBuff + offset, newESL, eslSize);
+       if (eslSize)
+               memcpy(*outBuff + offset, newESL, eslSize);
```


aaaand I just wrote all this and see there is an entirely different PR for fixes...
So we just need to remember to come back to documentation and cleaning up dead code in the the test suite
Nice work here! Thanks.

nick-child-ibm:
Please rebase onto latest `guest-devel` and I will merge

nick-child-ibm:
build fails due to `defined_sb_variables` being uint8_t and passed into strcmp:
```
backends/guest/common/util.c: In function ‘is_secure_boot_variable’:
backends/guest/common/util.c:180:34: error: pointer targets in passing argument 1 of ‘strcmp’ differ in signedness [-Werror=pointer-sign]
   if (strcmp(defined_sb_variables[i], var) == 0)
              ~~~~~~~~~~~~~~~~~~~~^~~
In file included from backends/guest/common/util.c:6:
/usr/include/string.h:137:12: note: expected ‘const char *’ but argument is of type ‘const uint8_t *’ {aka ‘const unsigned char *’}
 extern int strcmp (const char *__s1, const char *__s2)
            ^~~~~~

backends/guest/guest_svc_read.c: In function ‘read_path’:
backends/guest/guest_svc_read.c:284:37: error: pointer targets in passing argument 1 of ‘strcmp’ differ in signedness [-Werror=pointer-sign]
          strcmp(defined_sb_variables[i], PK_VARIABLE) == 0))
                 ~~~~~~~~~~~~~~~~~~~~^~~
In file included from backends/guest/guest_svc_read.c:11:
/usr/include/string.h:137:12: note: expected ‘const char *’ but argument is of type ‘const uint8_t *’ {aka ‘const unsigned char *’}
 extern int strcmp (const char *__s1, const char *__s2)
            ^~~~~~
backends/guest/guest_svc_read.c:289:33: error: pointer targets in passing argument 3 of ‘print_variables’ differ in signedness [-Werror=pointer-sign]
             defined_sb_variables[i]);
             ~~~~~~~~~~~~~~~~~~~~^~~
In file included from backends/guest/guest_svc_read.c:17:
./backends/guest/include/common/read.h:64:5: note: expected ‘const char *’ but argument is of type ‘const uint8_t *’ {aka ‘const unsigned char *’}
 int print_variables(const uint8_t *buffer, size_t buffer_size, const char *var_name);
     ^~~~~~~~~~~~~~~
```

With the following, build finishes and all tests are passing:
```
diff --git a/backends/guest/common/util.c b/backends/guest/common/util.c
index 9a25198..faef127 100644
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -177,7 +177,7 @@ bool is_secure_boot_variable(const char *var)
        int i = 0;
 
        for (i = 0; i < defined_sb_variable_len; i++) {
-               if (strcmp(defined_sb_variables[i], var) == 0)
+               if (strcmp( (const char *) defined_sb_variables[i], var) == 0)
                        return true;
        }
```
I have libc version 2.28 btw.

test/Makefile
erichte-ibm: Slated for a later rework of the tests. Most of the "helper" code for the tests can be factored out into shared classes, will tackle that soon.

Makefile
erichte-ibm: Indeed. I'll overhaul the build / testing instructions after the build system changes stabilize.

erichte-ibm:
> Please rebase onto latest `guest-devel` and I will merge

I fully bet that GitHub no longer things these histories are related because of the merge, and that is incredibly frustrating that I now will have to do this for *every PR*. Awesome.

PR #48, opened 2023-08-08
Reimplement CMake-based build system

erichte-ibm:
The original CMake build implementation allowed for easier packaging, so for the same reason, re-implement the CMake build now that some of the architectural problems with guest/host code has been sorted out.

The goals for this implementation are as follows:
 - provide an alternative for simple builds to the Makefile
 - easier integration into packaging systems (e.g. rpmbuild)
 
 This is *NOT* intended to be feature-parity with the Makefile, as (at the moment), running tests and other pseudo-targets are much easier done with simple make rules.

Marking as a draft for now, since I'm assuming there's features missing still. Please take a look and see if there are any critical missing features.

At the moment, the following features are not implemented, and may not be for the scope of this PR:
- Crypto backend selection (same as Makefile, TBD)
- Guest/host backend selection
- Running test cases, or anything involving tests
- Pseudo build targets like `cppcheck`, `format`, etc

### DIFF
--- a/.github/workflows/x86_CI.yml
+++ b/.github/workflows/x86_CI.yml
@@ -13,7 +13,7 @@ jobs:
         submodules: 'recursive'
 
     - name: install openssl and valgrind for running test script
-      run:  sudo apt-get update && sudo apt-get install -y openssl valgrind libssl-dev libmbedtls-dev cppcheck
+      run:  sudo apt-get update && sudo apt-get install -y openssl valgrind libssl-dev libmbedtls-dev cppcheck cmake
 
     - name: run cppcheck
       run: make cppcheck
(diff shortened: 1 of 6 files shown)


nick-child-ibm:
Build is working and passes test. Can't think of anything else it needs other than what is in your list. At this point, non essesntial build targets like cppcheck, generate tests and  format don't really need to be mirrored to CMake at all. For now, lets just think in terms of what could be useful when packaging. In those circumstances, I really only would like to see crypto and guest/host selection but that seems dependent on libstb-secvar and still needs Makefile implementation. So I would be fine merging this. LMK

nick-child-ibm:
Also reimplement CI checks for cmake builds

PR #50, opened 2023-08-18
More build tweaks, single backend edition

erichte-ibm:
This PR is not based on #49 , and may have minor conflicts. Since this one is much smaller, consider reviewing and merging this one first.

This PR reworks the build system to support selecting which backend to use, and which crypto library to link against.

Major changes:
 - split backend-specific dependencies into their respective makefile/cmake include files
 - remove unneeded files for a build, such as the extra mbedtls pkcs7
 - start the process of splitting the crypto implementations, though that will really depend on the forthcoming crypto library rework
 - (Makefile): replace the `OPENSSL=1`, `MBEDTLS=0`, etc options with a single `CRYPTO=openssl` variable
 
 ---

## Future work

Now that single backends have been implemented, and crypto library selection "in place": this is a good opportunity to really flex the CI. I will be spinning up a new branch to rework the host/guest tests to only run if the built binary supports host/guest, and then create a test matrix for guest-only, host-only, both -- for each crypto library. Actions will be split from cmake build and makefile build independently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,39 +8,37 @@ add_executable( secvarctl )
 
 target_sources( secvarctl PRIVATE secvarctl.c generic.c )
 
-set( INCLUDE_PATHS
+target_sources( secvarctl PRIVATE
+  # external/extraMbedtls/pkcs7.c
+  # external/extraMbedtls/pkcs7_write.c
+  # external/skiboot/libstb/secvar/crypto/crypto-mbedtls.c
+  external/skiboot/libstb/secvar/crypto/crypto-openssl.c
+  # external/skiboot/libstb/secvar/crypto/crypto-gnutls.c
+)
+
+
+target_include_directories( secvarctl BEFORE PRIVATE
   include/
   external/libstb-secvar/
   external/libstb-secvar/include/
   external/libstb-secvar/include/secvar/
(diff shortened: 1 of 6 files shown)


nick-child-ibm:
The only weird thing with this is `make HOST_BACKEND=1` will still build with both backends, in other words, single backend builds requires turning the other backend off. This is fine, the functionality is there and that is what matters. Additonally, runnning `make HOST_BACKEND=0 check` should only run the guest test suite but both are run. I know there is a test PR open so going to merge this and we can fixup the testing issue later on (or maybe #49 fixes this iddkk). Thanks again!

erichte-ibm:
> The only weird thing with this is make HOST_BACKEND=1 will still build with both backends, in other words, single backend builds requires turning the other backend off

I agree this is kind of weird, but I think I'll write it in the forthcoming `README` update as a "disable" option, so instead you would do `GUEST_BACKEND=0` if you wanted host-only. Kinda weird still though.

> Additonally, runnning make HOST_BACKEND=0 check should only run the guest test suite but both are run

Confirming that indeed, that's next on my list of reworks -- checking backend availability in the tests and skipping tests for unavailable backends. Bonus points for doing this as a build matrix in actions so all combos are tested.

PR #51, opened 2023-08-23
Use libstb-secvars crypto

nick-child-ibm:
So I removed the crypto.h and crypto-openssl.c file in skiboot (which we were using before), I left the gnutls and mbedtls files there for later porting (openssl is still the only supported lib atm).

We will now compile with the crypto implementation defined in libstb-secvar. Everything is more or less the same minus the removal of some wrappers that guest was using to access the crypto functions. So all of this work is just reacting to those subtle changes.

This work depends on the recent changes to libstb-secvars crypto work:
branch : https://github.com/nick-child-ibm/libstb-secvar/tree/nick_secvarctl_crypto
libstb-secvar PR: [PR 11](https://github.com/IBM/libstb-secvar/pull/11)


Makefile
```diff
@@ -39,18 +39,20 @@ EXTERNAL_SRCS =
 ifeq ($(CRYPTO), openssl)
   _LDFLAGS += -lcrypto
   _CFLAGS += -DSECVAR_CRYPTO_OPENSSL
-  EXTERNAL_SRCS += external/skiboot/libstb/secvar/crypto/crypto-openssl.c
+  EXTERNAL_SRCS += external/libstb-secvar/src/crypto_openssl.c
```
erichte-ibm, line 42: I think this should be included by default now as part of `libstb-secvar-openssl.a`, so we might not need to include external `*.c` files at all for crypto once all are ported over.

erichte-ibm:
I'm not sure what is going on with the submodule commit in [1d4eb4f](https://github.com/open-power/secvarctl/pull/51/commits/1d4eb4f2a4abfe0549d7bc4d5b1421b2455289af) , but I merged https://github.com/IBM/libstb-secvar/pull/11 so that commit should be replaced with the new head of the `secvarctl-devel` branch.

backends/guest/common/generate.c
```diff
@@ -449,35 +450,38 @@ int is_x509certificate(const uint8_t *buffer, const size_t buffer_size, uint8_t
...
-		*hash_funct = x509_hash_func;
+		hash_funct = x509_hash_func;
 	} else {
 		data_size = buffer_size;
 		data = (uint8_t *)buffer;
 	}
 
-	rc = crypto.generate_md_hash(data, data_size, (*hash_funct)->crypto_md_funct, &hash_data,
+	rc = crypto_md_generate_hash(data, data_size, get_crypto_alg_id(x509_hash_func), &hash_data,
 				     hash_data_size);
```
erichte-ibm, line 481: `x509_hash_func` is left uninitialized if `is_x509certificate` returns `!SUCCESS`, and that uninitialized value is passed into `crypto_md_generate_hash()`. Furthermore, `hash_funct` is assigned to in the `true` case of that if, and not referred to again.

Judging from my short understanding, I think the function at the end should be using `hash_funct` instead of `x509_hash_func`? e.g.
```C
rc = crypto_md_generate_hash(data, data_size, get_crypto_alg_id(hash_funct), &hash_data,
                             hash_data_size);
```

PR #52, opened 2023-08-25
Add secvarctl.spec in order to build RPMs

gcwilson:
Here's a spec file that successfully builds and packages secvarctl.  I've tested it on RHEL 8.7 and it should work on RHEL 9 and Fedora.

### DIFF
--- a/secvarctl.spec
+++ b/secvarctl.spec
@@ -1,0 +1,40 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2023 IBM Corp.
+Name:		secvarctl
+Version:        1.0
+Release:        1%{?dist}
+Summary:        PowerPC secure variable control
+BuildArch:      ppc64le ppcl64 x86_64
+
+Group:          None
+License:        Apache 2.0
+URL:            https://github.com/open-power/secvarctl
+Source0:        https://github.com/open-power/secvarctl/archive/v%{version}.tar.gz#/%{name}-%{version}.tar.gz
+
+BuildRequires: cmake openssl-devel
+Requires: openssl
+
+%description
+%{name} facilitates manipulation of PowerPC authenticated variables from
+userspace via sysfs.  The primary use case is management of secure boot
+variables.
+
+%global debug_package %{nil}
+
+%prep
+%setup -q
+%cmake
+
+%build
+%cmake_build
+
+%install
+%cmake_install
+  
+%files
+%defattr(-,root,root)
+%license LICENSE
+%doc README.md
+%{_mandir}/man1/%{name}.1.gz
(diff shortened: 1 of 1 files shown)


nick-child-ibm:
Thanks George, 
I compared against the currently deployed spec files for [fedora](https://src.fedoraproject.org/rpms/secvarctl/blob/rawhide/f/secvarctl.spec) and [suse](https://build.opensuse.org/package/view_file/openSUSE:Factory/secvarctl/secvarctl.spec?expand=1), everything looks the same but we may need to add a dependency for openssl.

gcwilson:
Added BuildRequires: openssl-devel and Requires: openssl plus bonus copyright statement and whitespace fixes.

secvarctl.spec
```diff
@@ -0,0 +1,40 @@
...
+%description
+%{name} facilitates manipulation of PowerPC authenticated variables from
+userspace via sysfs.  The primary use case is management of secure boot
+variables.
+
+%global debug_package %{nil}
+
+%prep
+%setup -q
+cmake . -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="/usr"
```
erichte-ibm, line 27: This can just be `%cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="/usr"`, later versions of rpmbuild in `el9`, fedora, etc actually set the build output directory, which is expected in the later `%cmake_build` call.

secvarctl.spec
```diff
@@ -0,0 +1,40 @@
...
+
+%build
+%cmake_build .
+
+%install
+%cmake_install .
```
erichte-ibm, line 33: Similarly, these don't need the path set; can remove the trailing `.`.

erichte-ibm (changes requested):
Looks good, I marked two places where there need to be slight tweaks to better use the `%cmake` functions for newer distros to build.

Commit history is a mess in this branch, but I have a working version [here](https://github.com/erichte-ibm/secvarctl/blob/erichte/experimental-actions/secvarctl.spec) and build output on multiple distros [here](https://github.com/erichte-ibm/secvarctl/actions/runs/6030937474).

secvarctl.spec
erichte-ibm: Quick update because I just thought to look at the [documentation](https://docs.fedoraproject.org/en-US/packaging-guidelines/CMake/), but it looks like `-DCMAKE_INSTALL_PREFIX` is also set by default by invoking `%cmake`, so only the build type setting is needed. I will test this tomorrow.

erichte-ibm (approved):
Looks good to me, thanks for the updated version!

Issue #58, opened 2023-10-02
make install does not work in 1.0.0-rc1


hramrach:
[build.log](https://github.com/open-power/secvarctl/files/12783024/build.log)

Looks like the path of the file to install is bad

PR #59, opened 2023-10-02
Makefile: Fix installation after source reorganization

hramrach:
Fixes: #58


nick-child-ibm:
Thanks Michal. Sorry you bumped into that. I'll add a goal to add a `make install` test in our CI

hramrach:
The other option is to not call cmake support experimental and drop support for plain makefiles.

Supporting two build systems is not so great.

PR #66, opened 2023-10-04
guest/generate: change --append option to be a boolean based on presence

### LINKED ISSUE #65: v1.0.0.-rc1 - -a option is accepting strings
Failing scenarios:
```
[root@ltcrain80-lp2 secvarctl]# /usr/bin/secvarctl -m guest generate c:a -k ./test/guest/testdata/goldenkeys/PK/PK.key -c ./test/guest/testdata/goldenkeys/PK/PK.crt -n PK -i ./test/guest/testdata/goldenkeys/PK/PK.crt -o pk.auth -a %d234324
RESULT: SUCCESS
[root@ltcrain80-lp2 secvarctl]# /usr/bin/secvarctl -m guest generate c:a -k ./test/guest/testdata/goldenkeys/PK/PK.key -c ./test/guest/testdata/goldenkeys/PK/PK.crt -n PK -i ./test/guest/testdata/goldenkeys/PK/PK.crt -o pk.auth -a string1
RESULT: SUCCESS
```

erichte-ibm:
Fixes #65.

Currently, the -a --append option accepts a string to be parsed into an integer, either 0 or 1. This integer is then used to set a boolean append flag.

Avoid parsing strings at all by having the boolean append flag be set to 1 if -a/--append exists, otherwise default to 0.

NOTE: This does introduce a change in the command line API.

---

This one is a slightly more invasive change, but I think the removal of needless string parsing is better for overall safety. Please let me know if you disagree with this approach.

### DIFF
--- a/backends/guest/guest_svc_generate.c
+++ b/backends/guest/guest_svc_generate.c
@@ -477,12 +477,7 @@ static int parse_options(int key, char *arg, struct argp_state *state)
 		}
 		break;
 	case 'a':
-		args->append_flag = strtol(arg, NULL, 0);
-		if (args->append_flag < 0 || args->append_flag > 1) {
-			prlog(PR_ERR, "ERROR: append flag accept 0 or 1 only but it is %d\n",
-			      args->append_flag);
-			rc = ARG_PARSE_FAIL;
-		}
+		args->append_flag = 1;
 		break;
 	case ARGP_KEY_ARG:
 		/* check if reset key is desired */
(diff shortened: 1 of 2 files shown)


nick-child-ibm:
I think the API breakage is worth the fix. I agree with your argument.

Though we really need to add a test for this flag being properly applied in our test suite.
I WAS able to manually run this patch and use `validate` to observe that it does properly detect the presence of the append flag.
So the patch is working and we can delay the addition of a test case until we get around to increasing code coverage.

As always, thank you!

Issue #61, opened 2023-10-03
v1.0.0-rc1 - Multiple ESLs can't read


nasastry:
While reading the grubdb: it has 6 ESL’s but secvarctl read only one and thrown error.

```
...
                85:42:F6:AF:EE:9C:10:2D:47:18:5D:B8:09:66:09:CF:72:00:6B:F7

ERROR: invalid ESL size (184844848)
        Found 1 ESL's

RESULT: SUCCESS
```

with the internal secvarctl (probable git head at e3658f2ce5d0089e72eb243e8deacaa2ddd577a4) it was showing all 6 ESLs

```
...
	Found 6 ESL's

RESULT: SUCCESS
```

nasastry:
Couldn't read the SBAT secvar:
```
# secvarctl read -n sbat
READING sbat :
	ESL SIG LIST SIZE: 51
	GUID is : 50ab5d6046e00043abb63dd810dd8b23
	Signature type is: SBAT
	Data: sbat,1


	DELETE-MSG: sbat,1


ERROR: invalid signature type
	Found 0 ESL's

RESULT: SUCCESS
```

with internal secvarctl could read SBAT:
```
# /home/secvarctl/secvarctl -m guest read -n sbat
READING sbat :
	Timestamp: 0000-00-00 00:00:00 UTC
	ESL SIG LIST SIZE: 51
	GUID is : 50ab5d6046e00043abb63dd810dd8b23
	Signature type is: SBAT
	Data: sbat,1


	Found 1 ESL's

RESULT: SUCCESS
```

nasastry:
with RC2 could read all grubdb and sbat
```
[root@ltcrain80-lp2 home]# secvarctl read -n sbat
READING sbat :
ESL 1:
	ESL SIG LIST SIZE: 51
	GUID is : 50ab5d6046e00043abb63dd810dd8b23
	Signature type is: SBAT
	Data: sbat,1


	Found 1 ESL's

RESULT: SUCCESS
[root@ltcrain80-lp2 home]# secvarctl read -n grubdb
READING grubdb :
ESL 1:
	ESL SIG LIST SIZE: 1083
	GUID is : a159c0a5e494a74a87b5ab155c2bf072
	Signature type is: X509
	Certificate-1: 	Found certificate info
 Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
[...]

PR #67, opened 2023-10-05
guest/read: rewrite print_variables function to use libstb-secvar helper functions

### LINKED ISSUE #63: v1.0.0.-rc1 - secvarctl sigsegv while reading fuzzed ESL file
```
[root@ltcrain80-lp2 home]# secvarctl read -e PK.esl.sigsegv
	ESL SIG LIST SIZE: 857
	GUID is : a159c0a5e494a74a87b5ab155c2bf072
	Signature type is: X509
Segmentation fault (core dumped)
```

and with earlier secvarctl not seen the SIGSEGV - (probable git head at e3658f2ce5d0089e72eb243e8deacaa2ddd577a4)

```
[root@ltcrain80-lp2 home]# /home/secvarctl/secvarctl -m guest read -e PK.esl.sigsegv
	ESL SIG LIST SIZE: 857
	GUID is : a159c0a5e494a74a87b5ab155c2bf072
	Signature type is: X509
	Found 0 ESL's

RESULT: SUCCESS

```

erichte-ibm:
Fixes #63, and maybe #61.

As reported in #63, a fuzzed ESL file causes a segfault when reading. This occurs because the fuzzed ESL contains an internal size value that is far larger than that of the ESL file itself. Therefore, when we hand the data to OpenSSL to parse, we give the parsing function a very incorrect size value to expect, and therefore it overruns the buffer.

Rather than add in more size checks, the function has been rewritten to use the ESL/ESD iteration helper functions in libstb-secvar, which already have coverage testing.

---

### Notes

I *think* this mostly maintains output parity with that of the older version, it appears to be the same for the test files that we have in the repo. I'm not sure about output from ESLs containing hashes -- we don't seem to have any of those in our testdata.

As a side-effect of rewriting that function, it now seems to read two appended ESLs without issue (e.g. testing on a file created like so: `cat PK.esl db.esl > two.esl`). I'm not positive on this though, so this could use some extended testing.

There's a few code-related changes I'm not positive on, I will make inline comments for those.

Edit: added a test case and included the invalid ESL. We should definitely consider either including more of these tests, or perhaps reuse some of the test data from libstb-secvar.


backends/guest/common/read.c
```diff
@@ -121,59 +122,46 @@ int print_cert_info(crypto_x509_t *x509)
...
 			case ST_HASHES_START ... ST_X509_HASHES_END:
 				printf("\tData-%zu: ", count);
-				print_hex(cert, cert_size);
+				print_hex(curr_esd.esd->signature_data, esd_data_size);
 			case ST_X509:
-				x509 = crypto_x509_parse_der(cert, cert_size);
+				x509 = crypto_x509_parse_der(curr_esd.raw, esd_data_size);
 				if (!x509)
 					break;
 				printf("\tCertificate-%zu: ", count);
```
erichte-ibm, line 167: The count used in the print here is the count for the ESLs. Is this intended?

I'm thinking of the case reading from a file that contains ESL A and ESL B, where ESL A contains 1 ESD, and ESL B contains 2 ESDs. This will print something like:

```
Certificate-0:
Certificate-1:
Certificate-1:
```

I feel like this should probably iterate per ESD, and maybe add an ESL separator with the ESL number?

backends/guest/common/read.c
```diff
@@ -121,59 +122,46 @@ int print_cert_info(crypto_x509_t *x509)
...
-			data_size = signature_size - sizeof(sv_esd_t);
-			cert = esd->signature_data;
-			cert_size = data_size;
-
+		while (curr_esd.esd != NULL) {
 			switch (sig_type) {
 			case ST_HASHES_START ... ST_X509_HASHES_END:
 				printf("\tData-%zu: ", count);
-				print_hex(cert, cert_size);
+				print_hex(curr_esd.esd->signature_data, esd_data_size);
```
erichte-ibm, line 162: Not sure if this should be the `signature_data` printing here or just the whole ESD, but I have no files to test with to get expected output.

backends/guest/common/read.c
```diff
@@ -186,30 +174,31 @@ int print_variables(const uint8_t *buffer, size_t buffer_size, const char *var_n
...
 			case ST_DELETE:
 				printf("\tDELETE-MSG: ");
-				print_raw((char *)cert, cert_size);
+				print_raw((char *)curr_esd.raw, esd_data_size);
+				break;
 			default:
-				prlog(PR_ERR, "ERROR: invalid signature type\n");
-				goto outside;
+				prlog(PR_ERR, "ERROR: invalid signature type = %d\n", sig_type);
+				break;
```
erichte-ibm, line 185: 99.9% sure was a bug, I don't know why we would want to fallthrough from `SBAT` into `DELETE` into `default`

backends/guest/common/read.c
nick-child-ibm: Wanted to test how this goes but it turns out `test/guest/testenv/dbx/data` is an ESL with a cert in it:
```
[nchild@tinkpad secvarctl]$ ./bin/secvarctl -m guest read -p test/guest/testenv/ -n dbx 
READING dbx :
	ESL SIG LIST SIZE: 857
	GUID is : a159c0a5e494a74a87b5ab155c2bf072
	Signature type is: X509
	Certificate-1: 	Found certificate info
 Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            ea:4d:46:f1:fd:93:30:f3:54:ec:dd:02:19:00:54:3d:3f:1e:cb
        Signature Algorithm: sha256WithRSAEncryption
        Issuer:
            countryName               = NC
            organizationName          = testing corp
        Validity
            Not Before: Apr  7 09:18:56 2023 GMT
            Not After : May  7 09:18:56 2023 GMT
        Subject:
            countryName               = NC
            organizationName          = testing corp
        X509v3 extensions:
            X509v3 Subject Key Identifier: 
                61:3A:96:F3:36:CF:40:53:46:25:38:49:66:F3:CF:BF:38:18:C9:1F
            X509v3 Authority Key Identifier: 
                keyid:61:3A:96:F3:36:CF:40:53:46:25:38:49:66:F3:CF:BF:38:18:C9:1F

            X509v3 Basic Constraints: critical
                CA:TRUE
 
	Found 1 ESL's

RESULT: SUCCESS
```
I am not 100% sure whats supported so whatever. 

I then went to go generated a esl with a hash and hit more bugs so opening a PR shortly. Anyways  for brevity jumping over that dilemma.

When I finally got a generated ESL w/ hash, I could test and looks like we are overreadding:
```
[nchild@tinkpad secvarctl]$ hexdump guest.dbx
0000000 1626 c1c4 504c 4092 a9ac f941 9336 2843
0000010 004c 0000 0000 0000 0030 0000 0000 0000
0000020 0000 0000 0000 0000 0000 0000 ea98 4f6e
0000030 6f21 b42f 9fb6 9bff 443a 2c84 6838 a66c
0000040 f385 5df5 8cc4 3f5d 10b1 e47b          
000004c

[nchild@tinkpad secvarctl]$ ./bin/secvarctl -m guest read -e guest.dbx
	ESL SIG LIST SIZE: 76
	GUID is : 2616c4c14c509240aca941f936934328
	Signature type is: SHA256
	Data-1: 38:68:6c:a6:85:f3:f5:5d:c4:8c:5d:3f:b1:10:7b:e4:00:00:00:00:00:00:00:00:00:00:00:00:11:04:00:00
	Found 1 ESL's

RESULT: SUCCESS

[nchild@tinkpad secvarctl]$ ./bin/secvarctl-dbg -m guest read -e guest.dbx
	ESL SIG LIST SIZE: 76
	GUID is : 2616c4c14c509240aca941f936934328
	Signature type is: SHA256
=================================================================
==60409==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60700000006c at pc 0x000000403413 bp 0x7ffe916eea50 sp 0x7ffe916eea40
READ of size 1 at 0x60700000006c thread T0
    #0 0x403412 in print_hex /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/generic.c:39
    #1 0x44fc00 in print_variables backends/guest/common/read.c:165
    #2 0x452a1b in read_esl backends/guest/guest_svc_read.c:156
    #3 0x456067 in guest_read_command backends/guest/guest_svc_read.c:452
    #4 0x4069b4 in main /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/secvarctl.c:249
    #5 0x7fb7c44d3d84 in __libc_start_main (/lib64/libc.so.6+0x3ad84)
    #6 0x4031dd in _start (/home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/bin/secvarctl-dbg+0x4031dd)

0x60700000006c is located 0 bytes to the right of 76-byte region [0x607000000020,0x60700000006c)
allocated by thread T0 here:
    #0 0x7fb7c5b59bb8 in __interceptor_malloc (/lib64/libasan.so.5+0xefbb8)
    #1 0x403b6e in get_data_from_file /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/generic.c:89
    #2 0x455d44 in guest_read_command backends/guest/guest_svc_read.c:440
    #3 0x4069b4 in main /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/secvarctl.c:249
    #4 0x7fb7c44d3d84 in __libc_start_main (/lib64/libc.so.6+0x3ad84)

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/generic.c:39 in print_hex
Shadow bytes around the buggy address:
  0x0c0e7fff7fb0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0e7fff7fc0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0e7fff7fd0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0e7fff7fe0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c0e7fff7ff0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c0e7fff8000: fa fa fa fa 00 00 00 00 00 00 00 00 00[04]fa fa
  0x0c0e7fff8010: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0e7fff8020: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0e7fff8030: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0e7fff8040: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0e7fff8050: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==60409==ABORTING
```
Host backend can parse this properly:
```
[nchild@tinkpad secvarctl]$ ./bin/secvarctl -m host validate -e guest.dbx -x -v
----opening guest.dbx is success: reading 76 bytes----
VALIDATING ESL:
	ESL SIG LIST SIZE: 76
	GUID is : 2616c4c14c509240aca941f936934328
	Signature type is: SHA256
	Hash: 98:ea:6e:4f:21:6f:2f:b4:b6:9f:ff:9b:3a:44:84:2c:38:68:6c:a6:85:f3:f5:5d:c4:8c:5d:3f:b1:10:7b:e4
	Found 1 ESL's

RESULT: SUCCESS
```
I was able to get the outputs to line up and stop ASAN complaints by making the following change:
```
diff --git a/backends/guest/common/read.c b/backends/guest/common/read.c
index be32ad8..1712e81 100644
--- a/backends/guest/common/read.c
+++ b/backends/guest/common/read.c
@@ -162,7 +162,7 @@ int print_variables(const uint8_t *buffer, size_t buffer_size, const char *var_n
                        switch (sig_type) {
                        case ST_HASHES_START ... ST_X509_HASHES_END:
                                printf("\tData-%zu: ", count);
-                               print_hex(curr_esd.esd->signature_data, esd_data_size);
+                               print_hex((unsigned char *)curr_esd.raw, esd_data_size);
                        case ST_X509:
                                x509 = crypto_x509_parse_der(curr_esd.raw, esd_data_size);
```
I needed to typecast because we are dropping the `const` to call `print_hex`, obviously you can opt to just modify the definition of `print_hex` to include const, also with this change all uses of `curr_edt` union only use one implementation `curr_edt.raw`

backends/guest/common/read.c
nick-child-ibm: Yeah i agree. Should be more like:
```
ESL 1:
  CERT 1:
  CERT 2:
 ESL 2:
   CERT 1:
```

backends/guest/common/read.c
```diff
@@ -186,30 +177,31 @@ int print_variables(const uint8_t *buffer, size_t buffer_size, const char *var_n
...
-
-		/* we read all esl_size bytes so iterate to next esl */
-		esl_data += next_esl_size;
-		esl_data_size -= next_esl_size;
 	}
 
 	printf("\tFound %zu ESL's\n\n", count);
 
-	if (x509)
-		crypto_x509_free(x509);
```
nick-child-ibm, line 212: Removing this exposes a possible leak if `print_cert_info` returns nonzero. Consider just adding a free call in the error block before the break.

nick-child-ibm (changes requested):
Thanks Eric, I think this looks a lot cleaner. Just minor fixes and your good to go

backends/guest/common/read.c
erichte-ibm: Do you prefer 0-indexing, or counting from 1 for each?

backends/guest/common/read.c
erichte-ibm: The arg to `print_hex` should probably be `const` anyway, so I will make that change. I'm not terribly married to the unions (pun...intended?) but I would like to overall reduce unnecessary casts in this code base.

backends/guest/common/read.c
```diff
@@ -121,59 +122,49 @@ int print_cert_info(crypto_x509_t *x509)
 int print_variables(const uint8_t *buffer, size_t buffer_size, const char *var_name)
```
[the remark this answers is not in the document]
erichte-ibm: So on this, I actually was originally going to split this into two functions, but opted for the easier option and just cleaned up this one instead. Want me to take care of both and split this into `print_esl_buffer` and `print_esd_buffer`? Else I'll just change the name.

backends/guest/common/read.c
nick-child-ibm: Doesn't mattrer to me just as long as the final message "Found %d ESLs" is correct

backends/guest/common/read.c
erichte-ibm: Fixed by always deallocating immediately after its only use. Related, return from the function if printing the cert fails. I don't really like the idea of limping along if anything fails to print.

backends/guest/common/read.c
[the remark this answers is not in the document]
erichte-ibm: I split it into two functions as a separate commit, can revert that commit if it is just needless factoring out.

erichte-ibm:
Pushed an updated branch, things to note:
 - Now prints both ESL # sequence and ESD # sequence as mentioned in comments. I think it's starting at 1, can change if we ever add operations that work on a specific ESL/ESD and 0-indexing is better.
 - changed `print_hex` to take in `const uint8_t` instead of non-`const` `unsigned char`
 - fixed possible `x509` memory leak by always deallocating after use.
 - changed ESD switch case range to use `ST_HASHES_END` instead of `ST_X509_HASHES_END` preemptively, not sure if this will cause or fix conflicts with the other PR
 - split and renamed `print_variables` to `print_esl_buffer` and (static helper) `print_esd_from_esl_buffer`
 - added an include to `prlog.h` in `guest/common/util.h` because something changed and had compiler errors, but that should be there anyway

backends/guest/common/read.c
```diff
@@ -111,6 +111,77 @@ int print_cert_info(crypto_x509_t *x509)
...
+			break;
+		default:
+			prlog(PR_ERR, "ERROR: invalid signature type = %d\n", sig_type);
+			break;
+		}
+
+		rc = next_esd_from_esl(esl, &curr_esd.raw, &esd_data_size, &esd_owner);
+		if (rc) {
+			prlog(PR_ERR, "Error reading next esd (%zu), rc = %d\n", esd_count, rc);
+		}
```
nick-child-ibm, line 179: We should probably break/return here or change the `while` conditional to check `rc`.
The loop will go as long as `curr_esd` is not `NULL`, there are several cases where `next_esd_from_esl` can return an errorcode and not update the return pointer to be NULL

erichte-ibm: Ah dangit, knew I missed one.

Also changing the error message for the default case nearby to be "unknown" signature type, since that's the one "error" we're allowing without exiting.

PR #70, opened 2023-10-05
Nick misc bugs

nick-child-ibm:
In essence, generating an ESL with hash contents was totally broken but returned success.

### DIFF
--- a/backends/guest/common/generate.c
+++ b/backends/guest/common/generate.c
@@ -436,6 +436,11 @@ int is_x509certificate(const uint8_t *buffer, const size_t buffer_size, uint8_t 
 
 	rc = validate_cert(cert, cert_size);
 
+	if (rc) {
+		free(cert);
+		return rc;
+	}
+
 	*cert_data = cert;
 	*cert_data_size = cert_size;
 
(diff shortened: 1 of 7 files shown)


erichte-ibm:
Thanks for removing the X509_HASHES definitions, had no idea what those were for either so I just ported exactly what the code was before.

Here's the two remaining uses of those definitions, I'll take a deeper look at the rest of the PR later today.
```
backends/guest/common/validate.c
88:	if (idx < ST_HASHES_END || idx < ST_X509_HASHES_START ||

backends/guest/common/read.c
172:			case ST_HASHES_START ... ST_X509_HASHES_END:
```

nick-child-ibm:
rebase to fix ci warnings. Also added another commit at the end for obvious conditional bug

backends/guest/common/generate.c
```diff
@@ -449,7 +454,7 @@ int is_x509certificate(const uint8_t *buffer, const size_t buffer_size, uint8_t
  * @param buffer_size , length of buffer
  * @param hash_funct, index of hash function information to use for ESL GUID,
  *                   also helps in prevalation, if inform is '[c]ert' then this doesn't matter
- * @param hash_data, the generated hash data
+ * @param hash_data, the generated hash data, buffer should be allocated before calling
```
erichte-ibm, line 457: Not worth going into for this PR, but on the topic of function names matching what they do: I don't like that a function implying boolean with `is_` was apparently allocating memory. Arguably this should be renamed to `try_`, or actually be made a boolean function that does not actually do any type conversion.

I'll add this to the backlog, but we need to make a pass on function names and behaviors.

erichte-ibm (approved):
LGTM, thanks for fixing this and the in-depth explanation!

PR #73, opened 2023-10-25
Type Usage Overhaul + Assorted Static Analysis Fixes

erichte-ibm:
## Type Overhaul

Buffer types are inconsistent across the project -- some bits use `char *`, `unsigned char *`, `uint8_t *`, etc. This PR attempts to unify all buffer types to use the same type to avoid needless casting.

The following has been done:
 - use `uint8_t *` for raw pointers to data (buffers)
 - use `char *` for strings and (hopefully) *only* strings
 - comment the few cases where a cast is necessary
   - this is primarily when loading data from a file that IS a string, or similar cases

---
## Static Analysis Stuff

Due to laziness on my part, I'm also piggybacking some static analysis fixes into this PR. These checks are not reported by the version of cppcheck that actions uses (2.7), but are included in the newer 2.12 version.

The changes are mostly minor:
 - ignore checking ccan headers, as we will not be amending those
 - use more `const`
 - scope reductions
 - suppress `unusedSuppression`, as cppcheck v2.12.1 didn't match some of the existing suppression. This should be reevaluated Later™
 - fix potential NULL reference in the PK/append special case

---
## On Skiboot changes

While I'd rather not amend external ccan headers, I did make some changes to the external skiboot code. While really that code should be reworked to work better as a library (and changes sent upstream), that is likely not going to happen. Thus, I have made some changes directly to the skiboot code to address more type issues and one static analysis fix. If we deem it necessary in the future, we should consider a full overhaul of the skiboot secure boot code (mostly reorganizing), and ship it to skiboot purely so that the files are consistent again. I will open a separate future-milestoned issue to keep this in mind, and it can be closed if we decide this isn't something we want to bother with.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,8 +155,9 @@ check: $(SECVAR_TOOL)
 
 CPPCHECK_FLAGS =  --enable=all --force -q --error-exitcode=1
 CPPCHECK_FLAGS += --suppress=missingIncludeSystem
-CPPCHECK_FLAGS += --suppress=unusedFunction       # false positive on validateTS
-CPPCHECK_FLAGS += --suppress=internalAstError     # false positive on ccan/list_for_each
+CPPCHECK_FLAGS += --suppress=unusedFunction             # false positive on validateTS
+CPPCHECK_FLAGS += --suppress=internalAstError           # false positive on ccan/list_for_each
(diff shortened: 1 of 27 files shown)


nick-child-ibm (approved):
All looks sane and safe. Thanks!

PR #75, opened 2023-10-26
Bug fixes

SudhakarKuppusamy1:
It has the following bug fixes
1. In verify/read, reading of timestamp from variable
2. In generate, trustedcadb allow only CA cert
3. misbehavior of verify with -p and write

### DIFF
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -38,6 +38,24 @@ const struct signature_type_info signature_type_list[] = {
 static uint8_t append[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 };
 static uint8_t replace[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
+void print_timestamp(timestamp_t t)
+{
+	printf("%04d-%02d-%02d %02d:%02d:%02d UTC\n", t.year, t.month, t.day, t.hour, t.minute,
+	       t.second);
+}
+
+void read_timestamp(const uint8_t *esl_data)
+{
+	timestamp_t timestamp;
+
+	if (esl_data == NULL)
+		return;
+
+	memcpy(&timestamp, esl_data + 1, TIMESTAMP_LEN - 1);
+	printf("\tTimestamp: ");
+	print_timestamp(timestamp);
+}
+
 /*
  * creates the append header using append flag
  */
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -38,6 +38,19 @@ const struct signature_type_info signature_type_list[] = {
 static uint8_t append[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 };
 static uint8_t replace[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
+/*
+ * check it whether given variable is trustedcadb
+ */
+bool is_trustedcadb_variable(const char *variable_name)
+{
+	int len = strlen(variable_name);
+
+	if (memcmp(variable_name, TRUSTEDCADB_VARIABLE, len) == 0)
+		return true;
+
+	return false;
+}
+
 void print_timestamp(timestamp_t t)
 {
 	printf("%04d-%02d-%02d %02d:%02d:%02d UTC\n", t.year, t.month, t.day, t.hour, t.minute,
--- a/backends/guest/common/verify.c
+++ b/backends/guest/common/verify.c
@@ -478,7 +478,7 @@ int validate_variables_arguments(struct verify_args *args)
 		return ARG_PARSE_FAIL;
 	}
 
-	if ((args->update_variable_size != 0 && args->variable_path == NULL) ||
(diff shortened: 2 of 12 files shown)


backends/guest/common/util.c
```diff
@@ -38,6 +38,19 @@ const struct signature_type_info signature_type_list[] = {
...
+bool is_trustedcadb_variable(const char *variable_name)
+{
+	int len = strlen(variable_name);
+
+	if (memcmp(variable_name, TRUSTEDCADB_VARIABLE, len) == 0)
+		return true;
+
+	return false;
+}
+
```
erichte-ibm, line 53:
```C
bool is_trustedcadb_variable(const char *variable_name)
{
    return !strcmp(variable_name, TRUSTEDCADB_VARIABLE);
}
```

backends/guest/common/verify.c
```diff
@@ -112,7 +114,7 @@ static int get_current_esl_data(const uint8_t *esl_file, uint8_t **current_esl_d
 			}
 		}
 	} else
-		return INVALID_FILE;
+		prlog(PR_WARNING, "WARNING: %s file does not have data\n", (char *)esl_file);
```
erichte-ibm, line 117: Why are we not returning an error code here? It seems like this can only be reached if we fail to allocate memory for the file, or fail to read from the file (`is_file` checks that it exists and can be opened, which *does* return an error code). In both cases, this should be signaled to the calling function that something went wrong.

erichte-ibm: Additional note: it appears `get_auth_data()` does return `INVALID_FILE` in its similar error `else` case, around ~L147

backends/guest/common/util.c
```diff
@@ -38,6 +38,37 @@ const struct signature_type_info signature_type_list[] = {
...
+void read_timestamp(const uint8_t *esl_data)
+{
+	timestamp_t timestamp;
+
+	if (esl_data == NULL)
+		return;
+
+	memcpy(&timestamp, esl_data + 1, TIMESTAMP_LEN - 1);
```
erichte-ibm, line 67: Why do we skip the first byte of the timestamp here?

Also, since this function does not do any check on the size of the buffer `*esl_data` points to, please include a comment that states that the caller must confirm the input buffer contains at least `TIMESTAMP_LEN` bytes.

backends/guest/guest_svc_read.c
```diff
@@ -261,12 +266,11 @@ static int read_path(const char *path, const int is_print_raw, const char *varia
...
 			if (is_print_raw || esl_data_size == DEFAULT_PK_LEN)
 				print_raw((char *)esl_data, esl_data_size);
-			else if (esl_data_size >= TIMESTAMP_LEN)
+			else if (esl_data_size >= TIMESTAMP_LEN) {
+				read_timestamp(esl_data);
 				rc = print_esl_buffer(esl_data + TIMESTAMP_LEN,
 						      esl_data_size - TIMESTAMP_LEN, variable_name);
-			else
-				prlog(PR_WARNING, "WARNING: The %s database is empty.\n",
-				      variable_name);
```
erichte-ibm, line 269: I'm a little unclear on the change to the conditionals here. I assume a variable will be completely empty if it hadn't yet stored data or freshly reset, but otherwise it should at least contain a data size of `TIMESTAMP_LEN`. Removing the else condition would remove some error cases that should handled.

It seems like there are four conditions:
1. `esl_data_size >= TIMESTAMP_LEN` ➔ normal case, variable data at least (probably) contains a timestamp
2. `0 < esl_data_size < TIMESTAMP_LEN` ➔ malformed variable contents?
3. `esl_data_size == 0` ➔ empty variable that hasn't been written to, might be worth printing that it is empty
4. `esl_data_size == -1` ➔ an error that *should* be printed, although I don't think this can be hit if `variable_from_path()` call returns `SUCCESS`.

erichte-ibm: Note this appears again later in the function.

erichte-ibm (changes requested):

backends/guest/common/util.c
soura15: From the structure of the plpks_get_variable under linux/arch/powerpc/platforms/pseries/plpks-secvar.c, it seems the esl_data size will always be fixed to align with timestamp_len, my concern is that do we have plans to support variable key sizes in the future? In such a case this will need refactoring...

backends/guest/common/verify.c
soura15: Acknowledged, we should return INVALID_FILE in this case as it will avoid unnecessary operations at a later stage.

backends/guest/guest_svc_read.c
soura15: I don't think conditions 1 and 4 can ever occur?  We can remove the esl_data_size != -1 check here.

backends/guest/common/util.c
soura15: This doesn't warrant a comment as per my understanding because we have used TIMESTAMP_LEN at multiple places and I don't think there can be a scenario where we will get esl_data_size less than the TIMESTAMP_LEN.

erichte-ibm:
Closing this PR, as these commits were merged in #77

Issue #76, opened 2023-11-17
guest: clean up and unify reading the timestamp from a variable buffer


erichte-ibm:
In multiple places in the code (namely `guest_svc_read.c:read_path()`), we load in data from a variable, then proceed to do some level of validation which mostly consists of checking that the buffer is an expected size before doing any parsing. As asked in PR #75 , the "expected" size of a buffer should be defined, documented, and probably abstracted out so that other places of the code can call a function to read in a particular variable and do this same validation.

---

From my understanding, the following conditions should be checked and handled on the buffer size:
 - `size == 0` → error case, likely invalid file or some other I/O error. variables should not be completely empty
 - `0 < size < HEADER_LEN` → error case, variable must contain at least the 16-byte header
 - `size == HEADER_LEN` → valid case, variable exists but contains no data
 - `size > HEADER_LEN` → valid case, data exists and can be parsed
 
Data read from a variable contains a header of 16 bytes, consisting of a 1-byte version number, then 15 bytes of timestamp, truncating the last pad so that the header remains aligned.

---

Cleaning this up will probably looking some like:
 - [ ] factor out header parsing into another function, perhaps `variable_from_path()`
 - [ ] remove references to `TIMESTAMP_LEN` and `TIMESTAMP_LEN - 1` when referring to this header
 - [ ] actually parse and inspect the version number?
 - [ ] clean up any duplicate code paths that differentiated reading from a variable vs reading from a local file

erichte-ibm:
Resolved in #77

PR #77, opened 2023-11-17
Add missing patches from internal version, redux

erichte-ibm:
Supercedes #75, can be closed when this is merged. This PR contains a set of amended commits, largely squashed down intermediate commits, and applied the fixes as suggested in the previous PR.

---

Notable changes from #75:
 - squashed unnecessary intermediate commits
 - added a comment about the version byte to `read_timestamp()`, further information and work can be tracked in #76 
 - restored `else` error cases that were removed for some reason
 - rebased onto `guest-devel`
 - restored returning `INVALID_FILE` from `guest_current_esl_data` in `common/verify.c`

### DIFF
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -45,17 +45,14 @@ void print_timestamp(timestamp_t t)
 	       t.second);
 }
 
-void read_timestamp(const uint8_t *esl_data)
+void read_timestamp(const struct signed_variable_header *data)
 {
 	timestamp_t timestamp;
 
-	if (esl_data == NULL)
+	if (data == NULL)
 		return;
 
-	// Special case: data read in from firmware contains a 16-bytes header, containing
-	//  a one-byte version number, then 15 bytes of timestamp -- truncating the trailing
-	//  padding byte at the end of the struct.
-	memcpy(&timestamp, esl_data + 1, TIMESTAMP_LEN - 1);
+	memcpy(&timestamp, &data->timestamp, sizeof(data->timestamp));
 	printf("\tTimestamp: ");
 	print_timestamp(timestamp);
 }
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -31,6 +31,14 @@ const struct signature_type_info signature_type_list[] = {
 static uint8_t append[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 };
 static uint8_t replace[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
+/*
+ * check it whether given variable is trustedcadb
+ */
+bool is_trustedcadb_variable(const char *variable_name)
+{
+	return !strcmp(variable_name, TRUSTEDCADB_VARIABLE);
+}
+
 void print_timestamp(timestamp_t t)
 {
 	printf("%04d-%02d-%02d %02d:%02d:%02d UTC\n", t.year, t.month, t.day, t.hour, t.minute,
--- a/backends/guest/common/util.c
+++ b/backends/guest/common/util.c
@@ -31,6 +31,27 @@ const struct signature_type_info signature_type_list[] = {
 static uint8_t append[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 };
 static uint8_t replace[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
+void print_timestamp(timestamp_t t)
+{
+	printf("%04d-%02d-%02d %02d:%02d:%02d UTC\n", t.year, t.month, t.day, t.hour, t.minute,
+	       t.second);
+}
+
+void read_timestamp(const uint8_t *esl_data)
+{
+	timestamp_t timestamp;
+
+	if (esl_data == NULL)
+		return;
+
+	// Special case: data read in from firmware contains a 16-bytes header, containing
+	//  a one-byte version number, then 15 bytes of timestamp -- truncating the trailing
+	//  padding byte at the end of the struct.
+	memcpy(&timestamp, esl_data + 1, TIMESTAMP_LEN - 1);
+	printf("\tTimestamp: ");
+	print_timestamp(timestamp);
+}
+
 /*
  * creates the append header using append flag
  */
--- a/backends/guest/common/verify.c
+++ b/backends/guest/common/verify.c
@@ -100,7 +100,8 @@ static int get_current_esl_data(const char *esl_file, uint8_t **current_esl_data
 	buffer = get_data_from_file(esl_file, SIZE_MAX, &buffer_size);
 	if (buffer != NULL && buffer_size >= TIMESTAMP_LEN) {
 		if (buffer_size == DEFAULT_PK_LEN) {
-			print_raw(buffer, buffer_size);
+			if (verbose >= PR_DEBUG)
+				print_raw(buffer, buffer_size);
 			free(buffer);
 			buffer = NULL;
 			buffer_size = 0;
--- a/backends/guest/common/verify.c
+++ b/backends/guest/common/verify.c
@@ -71,9 +71,10 @@ static int update_variable(const char *variable_name, const uint8_t *auth_data,
 		prlog(PR_INFO, "\tappend update: %s\n\n", (append_update ? "True" : "False"));
 
 		if (*new_esl_data != NULL) {
-			read_timestamp(*new_esl_data);
-			rc = print_esl_buffer((*new_esl_data + TIMESTAMP_LEN),
-					      (*new_esl_data_size - TIMESTAMP_LEN), variable_name);
+			read_timestamp((struct signed_variable_header *)*new_esl_data);
+			rc = print_esl_buffer((*new_esl_data + GUEST_HEADER_LEN),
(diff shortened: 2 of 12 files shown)


backends/guest/common/verify.c
```diff
@@ -95,13 +96,15 @@ static int get_current_esl_data(const char *esl_file, uint8_t **current_esl_data
...
 		if (buffer_size == DEFAULT_PK_LEN) {
 			print_raw(buffer, buffer_size);
 			free(buffer);
 			buffer = NULL;
 			buffer_size = 0;
-		} else if (buffer_size != TIMESTAMP_LEN) {
+		} else {
+			if (verbose >= PR_INFO)
+				read_timestamp(buffer);
 			rc = validate_esl(buffer + TIMESTAMP_LEN, buffer_size - TIMESTAMP_LEN);
```
nick-child-ibm, line 110: Seems our `TIMESTAMP_LEN` macro has a misleading name. Correct me if I am wrong but it is the len of the version + the timestamp?

Also `TIMESTAMP_LEN` is assigned the value of 8 (in `backends/guest/include/common/util.h`) but the definition of `timestamp_t` (in `external/libstb-secvar/external/edk2/common.h`) has  a length of 16?

erichte-ibm: You are correct, and it seems that `TIMESTAMP_LEN` is the length of how much of the timestamp that we actually use. I really don't like this, and I'll have to scan the code in another PR to see if we ever need to parse in the full 16 bytes (e.g. inscribed in an auth header, probably?). I'd rather use different types for parsing different data...

For now, I'm going to add comment to the definition of this constant, where its used, and `read_timestamp()`. I will lump the work to improve this into #76 .


Documenting here just in case it never gets documented properly:
```c
struct sv_timestamp
{
  leint16_t year;
  uint8_t month;
  uint8_t day;
  uint8_t hour;
  uint8_t minute;
  uint8_t second;  // <-- this is how much we actually read in with read_timestamp()
  uint8_t pad1;    // <-- TIMESTAMP_LEN stops here
  leint32_t nanosecond;  // \
  leint16_t timezone;    //  |-- unused section of the struct
  uint8_t daylight;      //  |
  uint8_t pad2;          // /
} SV_PACKED;
```

backends/guest/common/verify.c
```diff
@@ -95,13 +96,15 @@ static int get_current_esl_data(const char *esl_file, uint8_t **current_esl_data
 		return INVALID_FILE;
 
 	buffer = get_data_from_file(esl_file, SIZE_MAX, &buffer_size);
-	if (buffer != NULL) {
+	if (buffer != NULL && buffer_size >= TIMESTAMP_LEN) {
 		if (buffer_size == DEFAULT_PK_LEN) {
 			print_raw(buffer, buffer_size);
```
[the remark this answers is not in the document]
erichte-ibm: I don't know, I assume this is guest specific magic. I'll add a verbose check here though since regardless of *whatever* this is doing, it probably shouldn't just print stuff.

backends/guest/common/verify.c
erichte-ibm: > For now, I'm going to add comment to the definition of this constant, where its used, and read_timestamp(). I will lump the work to improve this into https://github.com/open-power/secvarctl/issues/76 .

JK new commit uses structs because it was easier to implement something better than document it.

erichte-ibm:
Commit 1021e33 is somewhat experimental, and I don't love the casts it introduces, but it shapes what I would rather have to resolve #76 . Let me know your thoughts on the implementation and names of the structs, I can either split that into its own PR, can continue to workshop it here, or if its good enough, just merge it here and I'll improve on it in a later PR.

backends/guest/common/verify.c
```diff
@@ -71,9 +71,10 @@ static int update_variable(const char *variable_name, const uint8_t *auth_data,
 		prlog(PR_INFO, "\tappend update: %s\n\n", (append_update ? "True" : "False"));
 
 		if (*new_esl_data != NULL) {
-			read_timestamp(*new_esl_data);
-			rc = print_esl_buffer((*new_esl_data + TIMESTAMP_LEN),
-					      (*new_esl_data_size - TIMESTAMP_LEN), variable_name);
+			read_timestamp((struct guest_variable_header *)*new_esl_data);
```
soura15, line 74: This is not a correct typecast.

backends/guest/include/common/util.h
```diff
@@ -52,6 +51,29 @@ struct signature_type_info {
...
+ */
+struct sv_trunc_timestamp {
+	leint16_t year;
+	uint8_t month;
+	uint8_t day;
+	uint8_t hour;
+	uint8_t minute;
+	uint8_t second;
+} __attribute__((packed));
+
```
soura15, line 65: This is an incomplete definition of the timestamp struct.

backends/guest/include/common/util.h
```diff
@@ -52,6 +51,29 @@ struct signature_type_info {
...
+/*
+ * 8-byte header at the beginning of any variable,
+ *  containing a version field, and a truncated timestamp
+ */
+struct guest_variable_header {
+	uint8_t version;
+	struct sv_trunc_timestamp timestamp;
+} __attribute__((packed));
```
soura15, line 73: This will give sizeof as 16?

soura15: Also considering we don't use the version anywhere, don't see the reason to maintain a separate struct for it?

backends/guest/include/common/util.h
soura15: Please reuse the timestamp_t struct here.

backends/guest/common/verify.c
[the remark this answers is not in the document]
soura15: Correct me if I am wrong, but since we are extracting the esl_data from the variable path, buffer_size will be greater than TIMESTAMP_LEN because it's a valid esl_file and the first 16 bytes are reserved for timestamp? So this check is not required?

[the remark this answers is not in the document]
xNickChildx: Generally yes, the current variable path is should only contain valid data but `secvarctl` also allows the user to pass in a custom path to treat as the current list of variables (this is useful for testing). At the end of the day, the check prevents possible buffer overread. Since we could be reading data provided by the user, I would prefer keeping this conditional

backends/guest/common/util.c
```diff
@@ -45,17 +45,14 @@ void print_timestamp(timestamp_t t)
...
 
-	if (esl_data == NULL)
+	if (data == NULL)
 		return;
 
-	// Special case: data read in from firmware contains a 16-bytes header, containing
-	//  a one-byte version number, then 15 bytes of timestamp -- truncating the trailing
-	//  padding byte at the end of the struct.
-	memcpy(&timestamp, esl_data + 1, TIMESTAMP_LEN - 1);
+	memcpy(&timestamp, &data->timestamp, sizeof(data->timestamp));
```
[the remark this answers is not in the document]
xNickChildx: > Why do we need a separate struct for the timestamp data?

We need a separate struct because timestamps are being used in 2 places:
1. When generating the auth file. This uses the full 16 byte timestamp_t structure. In the auth header is 16 bytes of timestamp, this value is also used in the hash. This auth structure we should try to keep consistent. These changes do not effect auth file generation.
2. When creating the exposed current secvar ESL. There is a 7 byte timestamp appended to the version right before the current esl. `libst-secvar` uses `unpack_signed_var` and `pack_signed_var` to cast between the 16 byte timestamp and 7 byte timestamp.

> Only changing the TIMESTAMP_LEN to 16 would suffice here.

If we read 16 bytes here then we would be loading in the first 9 bytes of esl data (9 bytes past the end of the timestamp) into the following `timestamp_t` values:
```
  uint8_t pad1;
  leint32_t nanosecond;
  leint16_t timezone;
  uint8_t daylight;
  uint8_t pad2;
  ```

File: backends/guest/include/common/util.h (hunk 4 above)
Comment (xNickChildx) [reply backends/guest/include/common/util.h:65]: Already noted why `timestamp_t` will not work here.

BUT checkout the top of `external/libstb-secvar/src/pseries.c`:
```
struct var_timestamp
{
  leint16_t year;
  uint8_t month;
  uint8_t day;
  uint8_t hour;
  uint8_t minute;
  uint8_t second;
} SV_PACKED;

typedef struct var_timestamp var_timestamp_t;

struct signed_variable
{
  uint8_t version; /* must be 0 */
  var_timestamp_t time;
} SV_PACKED;
```

I think we should work on exporting this struct to a headerfile in `libstb-secvar`. That way we won't need to maintain our own implementation. I am not sure what the current status is of guest-sb support or if FW is already dedicated to a release of libstb-secvar so let me know if we should keep this commit for now and revert some of it later  (after these structs are available in a header file in libstb-sevar) or we should just hold off on applying this commit.

xNickChildx (commented):
See comments.
Please advise on the possibility of having libstb-secvar host the sevar-esl-header-timestamp structure. If that will be a long process then we can merge once you, @soura15 and I are synced up and in agreement!
Thanks again and welcome @soura15 , great to have another reviewer!

backends/guest/include/common/util.h
erichte-ibm: I agree, and I will look into that. I don't like how there is just an undefined 8-byte header that comes from firmware, as that *should* be something in libstb-secvar for the exact reason of consistency between userspace and firmware.

It's up to you if you want to merge and take this now, or wait for the libstb-secvar change. We can always patch it out in a later release if need be.

PR #82, opened 2024-01-03
Fix generate argument parsing

### LINKED ISSUE #80: Does not support the append flag message printed 3 times
```
# /usr/bin/secvarctl -m guest generate c:a -k /home/secvarctl/test/guest/testdata/goldenkeys/PK/PK.key -c /home/secvarctl/test/guest/testdata/goldenkeys/PK/PK.crt -n PK -i /home/secvarctl/test/guest/testdata/goldenkeys/PK/PK.crt -o pk.auth -a 0
ERROR: PK does not support the append flag
ERROR: PK does not support the append flag
ERROR: PK does not support the append flag
RESULT: FAILURE
```

erichte-ibm:
Fixes #80 

Argp calls the `parse_options()` function for each argument, therefore the logic to reject the append flag on a PK was evaluated for other arguments. Move the append flag reject logic outside of the argument parser, since it only needs to be evaluated once anyway, after all arguments have been read.

---

While solving the above, I noticed that the positional format specifier argument parsing logic was fairly... lacking. It led to some confusing error messages, as the original test case was using the old `-a 0` syntax for specifying the append header. The `0` was treated as a positional argument, and therefore was expected to be in the `c:a` format. Since it was _not_, it overwrote the previously valid `c:a` format specifier with `NULL`s, leading to an error in the format specifier, NOT in an unexpected argument.

The second commit addresses this, by adding some immediate enforcement of the format specifier, returning an error if it does not fit, or if another positional argument is supplied.

### DIFF
--- a/backends/guest/guest_svc_generate.c
+++ b/backends/guest/guest_svc_generate.c
@@ -483,6 +483,19 @@ static int parse_options(int key, char *arg, struct argp_state *state)
 		args->append_flag = 1;
 		break;
 	case ARGP_KEY_ARG:
+		/* there should only be one format specifier, error if another is supplied */
+		if (args->input_form && args->output_form) {
+			prlog(PR_ERR, "ERROR: unknown additional positional argument %s\n", arg);
+			rc = ARG_PARSE_FAIL;
+			break;
+		}
+		/* both forms should be either set or NULL, this should never be reached. */
+		if (!args->input_form ^ !args->output_form) {
+			prlog(PR_ERR,
+			      "ERROR: only one of input_form/output_form is set, this should not happen\n");
+			rc = ARG_PARSE_FAIL;
+			break;
+		}
 		/* check if reset key is desired */
 		if (!strcmp(arg, "reset")) {
 			args->input_form = "reset";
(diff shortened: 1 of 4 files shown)


backends/guest/guest_svc_generate.c
```diff
@@ -493,14 +506,19 @@ static int parse_options(int key, char *arg, struct argp_state *state)
...
+			rc = ARG_PARSE_FAIL;
+		}
 		break;
 	case ARGP_KEY_SUCCESS:
 		/* check that all essential args are given and valid */
 		if (args->help_flag)
 			break;
-		else if (args->input_form == NULL || args->output_form == NULL)
-			prlog(PR_ERR, "ERROR: incorrect '<input_format>:<output_format>', see "
-				      "usage...\n");
```
nick-child-ibm, line 503: This final check is necessary otherwise the user can get away with not setting the `in:out` argument:
```
$ ./bin/secvarctl-dbg -m guest generate -i sbat_in -o sbat.auth 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==566168==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x0000004377a9 bp 0x7ffe8abfeae0 sp 0x7ffe8abfeab0 T0)
==566168==The signal is caused by a READ memory access.
==566168==Hint: address points to the zero page.
    #0 0x4377a8 in parse_options backends/guest/guest_svc_generate.c:525
    #1 0x7fe995198892 in argp_parse (/lib64/libc.so.6+0x13a892)
    #2 0x438dbc in guest_generate_command backends/guest/guest_svc_generate.c:666
    #3 0x4069b4 in main /home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/secvarctl.c:250
    #4 0x7fe995098d84 in __libc_start_main (/lib64/libc.so.6+0x3ad84)
    #5 0x4031dd in _start (/home/nchild/IBM/secvarctl-stuff/erichte-guest-devel/secvarctl/bin/secvarctl-dbg+0x4031dd)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV backends/guest/guest_svc_generate.c:525 in parse_options
==566168==ABORTING
```
This can avoided by adding back this conditional:
```
$ ./bin/secvarctl-dbg -m guest generate -i sbat_in -o sbat.auth 
ERROR: incorrect '<input_format>:<output_format>', see usage...
Usage: secvarctl -m guest generate [OPTION...]
            <input_format>:<output_format> -i <input_file> -o <output_file>
  or:  secvarctl -m guest generate [OPTION...]
            reset -i <input_file> -o <output_file>
Try `secvarctl -m guest generate --help' or `secvarctl -m guest generate
--usage' for more information.
failed during argument parsing
RESULT: FAILURE
```

erichte-ibm: Yep, that's an oversight on my part. Re-added the check, and tweaked the message to include potentially missing the argument.

erichte-ibm:
Updated:
 - re-added the check
 - added test cases to hopefully catch these in the future
 - added a new commit that allows the test framework to more properly catch exits via `SIGSEGV`, `SIGABRT`, etc

